PR #10, opened 2020-10-16
Remove-PnPTeamsUser - Update example 2

a1mery:
### What is in this Pull Request ?

Update of the second example to use the _Role_ parameter


erwinvanhunen:
good catch!

PR #64, opened 2020-12-18
Update Register-PnPAzureADApp.md

LeonArmston:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #X, partially fixes #Y, mentioned in #Z, etc.

Updated wording on -CertificatePath as it referenced password and not file path like it should.


erwinvanhunen:
good catch!

PR #68, opened 2020-12-27
Fix - Check the UniqueId column instead of GUID column

gautamdsheth:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
1) Mentioned in PR - https://github.com/pnp/PnP-PowerShell/pull/2344
2) Mentioned in Issue - https://github.com/pnp/PnP-PowerShell/issues/2343

Currently we filtering based on the GUID column instead of the UniqueId column. 
This PR fixes that. Used to cause confusion 
Please credit the original author if we merge this :)


erwinvanhunen:
We have to watch out with this one... it's quite a breaking change as such that is easily missed, especially in the context of scripts to be migrated. I'll give it some thought.

erwinvanhunen:
Why not modify the CAML query to use an or? E.g. check for both UniqueID or GUID value. As they're unique anyway, there will be no risk of conflicts that you find the wrong item.

gautamdsheth:
Haa :) , makes perfect sense.
Will modify the PR !

PR #69, opened 2020-12-27
Feature - added pipe binds to Add-PnPFile and Add-PnPFolder commands

gautamdsheth:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
NA

Added Folder pipe binds for Add-PnPFile and Add-PnPFolder commands.
Also, modified it to support folder names with special chars like `#` and `%` via resource path API.


erwinvanhunen:
Thank you!

PR #70, opened 2020-12-29
Added missing Force parameter for Remove-TermGroup

gautamdsheth:
## Type ##
- [x] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
NA

Added missing Force parameter for `Remove-PnPTermGroup` command.
Localization string message improved.
Added necessary changes in docs


erwinvanhunen:
While you can actually do -Confirm:$false on this cmdlet, in the line of the other cmdlets it makes sense to reintroduce the force parameter.

PR #78, opened 2021-01-04
Added a check for GetAvailableTagsForSite before applying label

### LINKED ISSUE #74: Set-PnPLabel does not check for GetAvailableTagsForSite
Set-PnPLabel does not check for [GetAvailableTagsForSite](https://docs.microsoft.com/en-us/dotnet/api/microsoft.sharepoint.client.compliancepolicy.sppolicystoreproxy.getavailabletagsforsite?view=sharepoint-csom). 

### Expected behavior 
When applying label from SharePoint site (List Settings> Apply Label), we only see the Labels that are available for the site. If a Label is passed that is not available for the site, we would expect an error stating the Label is not available for the site.
### Actual behavior 
Set-PnPLabel applies a Label that is not available for the specific site as long as the Label exists in Security and Compliance Center. This differs from the behavior in the site.

- Create a Retention Label Policy and publish one or more retention labels to that policy.
- Apply that Policy to a SharePoint site.
- Verify the labels from previous step are available in the site (dropdown in List Settings> Apply Label to this list/library)
- now, Set-PnPLabel -List $list -Label <Label that is not in the dropdown> -BlockDeletion $true BlockEdit $true SyncToItems $true
Latest. You can reproduce this issue in any versions of PnP.PowerShell or SharePointPnPPowerShellOnline.

- [x] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

bhishma:

## Type ##
- [x ] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #74 

Added a check for GetAvailableTagsForSite before applying label to a list or library.


erwinvanhunen:
Hi! Thank you for your PR. I see you catch the exceptions and turn them into warnings. I don't really agree with that approach as not catching the exception will allow script writers to handle the error  themselves in the script. That's the reason we don't catch exceptions in general in the Core SDK/PnP Framework/PnP PowerShell -unless- we are testing for something and the result of a failed test is an exception. If you can make that change I'm happy to merge it :-)

bhishma:
@erwinvanhunen, thank you for the comment. I'm a bit confused. I do see the 'current' version does catch exception and writes as warning if it fails to apply the label. Are you saying we shouldn't do that? I'll wait for your response and make the changes.

                  
                    var listUrl = list.RootFolder.ServerRelativeUrl;
                    Microsoft.SharePoint.Client.CompliancePolicy.SPPolicyStoreProxy.SetListComplianceTag(ClientContext, listUrl, Label, BlockDeletion, BlockEdit, SyncToItems);

                    try
                    {
                        ClientContext.ExecuteQueryRetry();
                    }
                    catch (System.Exception error)
                    {
                        WriteWarning(error.Message.ToString());
                    }

erwinvanhunen:
You're right! I missed that part. Indeed... us dumping the exception makes kinda more sense, however, it would change the behavior of the cmdlet. Never mind my comment.

PR #89, opened 2021-01-06
Update Get-PnPDiagnostics.md

LeonArmston:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

Fixed wording on Example 1 to be Get-PnPDiagnostics rather than Get-PnPDiagnostic and removed all the parameters (which were referencing Add-PnPApp) as there are no parameters for this command. Also adding out of the box PowerShell parameters as [<CommonParameters>] to show they can be used.


erwinvanhunen:
Thanks! Good catch!

PR #90, opened 2021-01-07
Update Remove-PnPTeamsTab.md

LeonArmston:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

Updated examples to remove reference to GroupId and ChannelId

Also tested commands and I note that -Team can accept Team display name and groupId  but -Channel can only accept channel display name and if you put in the channel id you get the error message below.

image


erwinvanhunen:
Hi! This was actually a bug, as it should work with the channelid too. I fixed it and it should now work. I updated the documentation too to add an extra sample.

LeonArmston:
Thanks - good to see you fixed it. This affected lots of the other PnP commands  for Teams when using channel ID so these should be fixed now.

PR #115, opened 2021-01-13
Cleanup Documentation - Deleted Commands

LeonArmston:
## Type ##
- [X] Documentation Updates
- [ ] New Feature
- [ ] Sample

Deleting Deprecated or Old (replaced) command documentation.


erwinvanhunen:
Thank you for these! Unfortunately the issue we had last weekend with the dev branch caused us to restore parts of it from the master branch, reintroducing these documentation files. Much appreciated that you're helping with the cleanup!

LeonArmston:
> Thank you for these! Unfortunately the issue we had last weekend with the dev branch caused us to restore parts of it from the master branch, reintroducing these documentation files. Much appreciated that you're helping with the cleanup!

Not a problem at all! I guessed it might have been something like that with the re-appearing of these files!

I've just created another pull request for the Unified Group commands that have come back to delete them.

Do let me know if there are any other areas of the documentation you believe may be incorrect and need checking?

PR #118, opened 2021-01-14
Bugfix in New-PnPTenantSite when using server relative URL

KoenZomers:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
N/A

Fixing a bug when using a server relative URL with `New-PnPTenantSite`, i.e.:

```powershell
New-PnPTenantSite -Title "BLANKINTERNET#0" -Url "/sites/BLANKINTERNET4" -Owner "admin@xxx.onmicrosoft.com" -Lcid 1033 -TimeZone 4
```

This would result in an exception.

Also removed the -Force option and the confirmation prompt as confirmation was being asked when providing a server relative URL but not when providing a full URL, which is inconsistent. The SPO Management Shell also doesn't prompt for confirmation so to align with its behavior, removing the entire prompt here as well.

### DIFF
--- a/src/Commands/Base/PnPAdminCmdlet.cs
+++ b/src/Commands/Base/PnPAdminCmdlet.cs
@@ -41,7 +41,6 @@ namespace PnP.PowerShell.Commands.Base
                 throw new InvalidOperationException(Resources.NoSharePointConnection);
             }
 
-
             PnPConnection.CurrentConnection.CacheContext();
 
             if (PnPConnection.CurrentConnection.TenantAdminUrl != null &&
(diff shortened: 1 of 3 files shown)


src/Commands/Base/PnPAdminCmdlet.cs
```diff
@@ -76,12 +73,9 @@ protected override void BeginProcessing()
                     PnPConnection.CurrentConnection.Context =
                         PnPConnection.CurrentConnection.CloneContext(adminUrl);
                 }
-                else if (PnPConnection.CurrentConnection.ConnectionType == ConnectionType.TenantAdmin)
```
KoenZomers, line 79: Removed this IF statement as the ConnectionType only knows 2 options: O365 and TenantAdmin. For both of these it should populate the _baseUrl in a similar fashion.

erwinvanhunen:
Removing the -Force parameter for instance will cause a breaking change. Better to mark the parameter as obsolete in the code and remove it from the documentation. Simply ignore it in the code then.

KoenZomers:
PR has been updated to accommodate the Force parameter remark.

erwinvanhunen:
Sorry to keep bugging you, but you can simply put an [Obsolete("message")] attribute over the Force parameter. That would allow us to clean up code easier in the future too as we can search for those attributes.

KoenZomers:
We'll get there.. some day.. :) Updated the PR

image

KoenZomers:
Crap now the rebase has made it look messy. Sigh.

KoenZomers:
Ok whew. Managed to clean it up again. Ready for a 3rd attempt :)

PR #129, opened 2021-01-19
Fix ParameterSet_ACSAPPONLY

sebastianmattar:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

Fixed parameter set for authentication scenarios with ACS, ClientId was not allowed.
This now works: Connect-PnPOnline -Url ... -ClientId ... -ClientSecret ...


erwinvanhunen:
Very good catch!

PR #148, opened 2021-01-22
[UPDATE] Support for OrgAssetType in Add-PnPOrgAssetsLibrary

YannickRe:
- Fixed docs to show correct default value for CdnType (Public)
- Added support for specifying the OrgAssetType that will be stored in the org asset library, default value "ImageDocumentLibrary". Brings it in line with the Add-SPOOrgAssetsLibrary


erwinvanhunen:
Thank you!

PR #204, opened 2021-02-02
Updated the registration command and added a step

derhallim:
## Type ##
- [x] Documentation Update

Without changing the allowPublicClient to true, users will get an error like this when trying to connect with credentials: 
AADSTS7000218: The request body must contain the following parameter: 'client_assertion' or 'client_secret'.

The solution is mentioned in step number 5 in this page: https://github.com/Azure-Samples/active-directory-dotnetcore-devicecodeflow-v2#register-the-client-app-active-directory-dotnet-deviceprofile

### DIFF
--- a/pages/articles/connecting.md
+++ b/pages/articles/connecting.md
@@ -27,12 +27,14 @@ Connect-PnPOnline -Url "https://[yourtenant].sharepoint.com" -Credentials (Get-C
 You will have to create your own Azure AD Application registration, or you can create one:
 
 ```powershell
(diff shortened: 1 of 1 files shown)


erwinvanhunen:
Thank you!  Maybe you can add a note that this only applies if people want to use the registration for credential based auth?

derhallim:
Yes will do!

derhallim:
@erwinvanhunen Done.

erwinvanhunen:
perfect! Thank you!

PR #213, opened 2021-02-04
Updated Register-PnPAzureADApp sample to use a result variable

KoenZomers:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [X] Sample

## Related Issues? ##
N/A

Updated Register-PnPAzureADApp sample to use a result variable for easier access to the base64 encoded output. Also added quotes around the name to show spaces are allowed in the naming.


erwinvanhunen:
If people run the example as written there, while it is smart to return the results into a variable, maybe you can add some description of why it is smart? Also if run as this by default they will not see the appid, e.g. they will have to investigate the result.

KoenZomers:
@erwinvanhunen Updated this PR with additional text clarifying why a result variable is suggested and a sample of how the base64 encoded pfx can be leveraged to connect.

Issue #157, opened 2021-01-23
[BUG] Fails to connect with Azure Automation


pkbullock:
When running PnP PowerShell in Azure Automation, using app-only and certificate, the latest version fails to connect with the error "An error occurred: The specified network password is not correct." when running latest modules. There is a sample that helps with setting this scenario up (detailed below) uses an older version of the module that work. I have rolled back and found a good working version where the issue started:

### Testing Modules

- pnp.powershell.0.2.24-nightly works
- pnp.powershell.0.3.35-nightly works
- pnp.powershell.0.3.36-nightly works
- **pnp.powershell.0.3.37-nightly fails**
- pnp.powershell.0.3.38-nightly fails
- pnp.powershell.1.1.0 fails
- pnp.powershell.1.2.0 onwards fails

The cmdlet that fails is the **Connect-PnPOnline**  with the following params when connecting in Azure Automation. After some testing, the fault was introduced in version pnp.powershell.0.3.37-nightly onwards.

The command with param combination in the runbook:

```powershell
$siteConn = Connect-PnPOnline -ClientId $appId -CertificateBase64Encoded $base64Cert `
        -CertificatePassword $azureAutomateCreds.Password `
        -Url $baseSite -Tenant $appAdTenant -ReturnConnection
```

### Expected behavior 
I think somewhere in the versions, something may have changed or the expected input format.

### Actual behavior 

![image](https://user-images.githubusercontent.com/8781041/105613476-9e88d580-5dba-11eb-8854-c4af474b9ef8.png)

### Steps to reproduce behavior

To help with to setup the environment, refer to the sample: https://github.com/pnp/powershell/tree/dev/samples/Connect.AzureAutomation
(note, there will be a PR coming soon to fix an issue on **Register-PnPAzureADApp** to add the parameter -DeviceLogin to get this to work)

After installing the sample, it should work because the sample currently uses version: pnp.powershell.0.2.24-nightly
Updating the module to latest, the connection fails with the error shown in the above screenshot.

### Which operating system/environment are you running PnP PowerShell on?
- [ ] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [X] Other : please specify: Azure Automation

pkbullock:
I found the answer, will submit a PR for adjusting the sample to the latest versions. In the runbook, I export the certificate to then encode to base 64, in the latest versions i need to supply the password with the export.

pkbullock:
[...]

PR #223, opened 2021-02-05
Updated Get-PnPAAdUser to allow retrieval of all users

KoenZomers:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
N/A

Updated `Get-PnPAAdUser` to allow retrieval of all users. The current implementation only returns the first 1000 users in an Azure Active Directory due to the default EndIndex of 999 in the underlying PnP Framework. Changed the implementation in PnP Framework through [PR #157](https://github.com/pnp/pnpframework/pull/157) to allow the EndIndex to be nullable in which case it will return all available results from Azure Active Directory. For backwards compatibility in PnP PowerShell when using `Get-PnPAadUser`, I have made it so that the current cmdlet still returns the first 1000 results and only if `-EndIndex $null` is provided it will return all users.

### DIFF
--- a/documentation/Get-PnPAADUser.md
+++ b/documentation/Get-PnPAADUser.md
@@ -21,7 +21,7 @@ Retrieves users from Azure Active Directory
 
 ### Return a list (Default)
 ```powershell
-Get-PnPAADUser [-Filter <String>] [-OrderBy <String>] [-Select <String[]>] [<CommonParameters>]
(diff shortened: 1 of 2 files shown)


KoenZomers:
@erwinvanhunen Since the documenation stated that it should return ALL users already but in reality it only would return the first 1000 users, perhaps its even better to rewrite this PR to just "fix" it instead of introducing -EndIndex $null that needs to be used to retrieve all users? Don't actually think fixing it would fall into the break backwards compatibility scenario. What do you think?

PR #243, opened 2021-02-09
Replace Thread.Sleep() with Task.Delay() in async methods

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
No referenced github issues

Replace Thread.Sleep() with Task.Delay() in async methods as Thread.Sleep() suspends the current thread that, in case of async methods, could be a thread-pool thread, not allowing it to process other tasks.


erwinvanhunen:
Good one! Thank you!

PR #249, opened 2021-02-09
Fix: delete cert from cache on disk

michael-jensen:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
None that I am aware of

When you run Connect-PnPOnline with parameters for App-Only with Azure Active Directory with a certificate file (i.e. using -CertificatePath) a temporary system file is written to C:\ProgramData\Microsoft\Crypto\RSA\MachineKeys and is never removed, even after your run the Disconnect-PnPOnline cmdlet.

The code to carry out this cleanup exists in the Disconnect-PnPOnline cmdlet, but it appears that a parameter was missed in the ConnectOnline.cs code to setup for this cleanup process, so this PR addresses this.

### DIFF
--- a/src/Commands/Base/ConnectOnline.cs
+++ b/src/Commands/Base/ConnectOnline.cs
@@ -408,7 +408,7 @@ namespace PnP.PowerShell.Commands.Base
                 {
                     ReuseAuthenticationManager();
                 }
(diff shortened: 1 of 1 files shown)


erwinvanhunen:
Thank you!

PR #250, opened 2021-02-10
Updates to Add/Get/RemoveEventReceiver cmdlets

### LINKED ISSUE #241: Add SiteCollection as scope to EventReceivers cmdlets
*-PnPEventReceiver cmdlets currently only support List or Web scopes. It would be wonderful to have Site as another scope to get site collection level event receivers.

```Get-PnPEventReceiver -Scope Site``` or something like that.

I currently use CSOM to work with site collection level event receivers.
```
$clientContext = Get-PnPContext
$site = $clientContext.Site; 
$clientContext.Load($site)
$eventReceivers = $site.EventReceivers
$clientContext.Load($eventReceivers)
Invoke-PNPQuery
```
**Additional context**

### PR DESCRIPTION
**Pull request by bhishma**

## Type ##
- [ ] Bug Fix
- [x ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #241

Updated Add/Get/RemoveEventReceiver cmdlets to allow Site Collection as an option. Currently these cmdlets only allow List or Web. 
It depends on an update to PnPFramework. See : https://github.com/pnp/pnpframework/pull/166

### DIFF
--- a/src/Commands/Base/PipeBinds/EventReceiverPipeBind.cs
+++ b/src/Commands/Base/PipeBinds/EventReceiverPipeBind.cs
@@ -45,6 +45,23 @@ namespace PnP.PowerShell.Commands.Base.PipeBinds
             return Name ?? Id.ToString();
         }
 
+        internal EventReceiverDefinition GetEventReceiverOnSite(Microsoft.SharePoint.Client.Site site)
+        {
+            if (_eventReceiverDefinition != null)
+            {
+                return _eventReceiverDefinition;
+            }
+
+            if (_id != Guid.Empty)
+            {
+                return site.GetEventReceiverById(_id);
+            }
+            else if (!string.IsNullOrEmpty(Name))
+            {
+                return site.GetEventReceiverByName(Name);
+            }
+            return null;
+        }
         internal EventReceiverDefinition GetEventReceiverOnWeb(Web web)
         {
             if (_eventReceiverDefinition != null)
--- a/src/Commands/Events/AddEventReceiver.cs
+++ b/src/Commands/Events/AddEventReceiver.cs
@@ -11,6 +11,9 @@ namespace PnP.PowerShell.Commands.Events
         [Parameter(Mandatory = false)]
         public ListPipeBind List;
 
+        [Parameter(Mandatory = false)]
+        public SitePipeBind Site;
+
         [Parameter(Mandatory = true)]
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by erwinvanhunen** [comment]
Hi!

Thank you for your work, but unfortunately there are some problems with this PR which prevents us from merging it. The GetEventreceiver.cs class for instance references to a ParameterSet called "Site", but it's not defined, so it will never resolve to that.

In the AddEventReceiver.cs you define a SitePipeBind class called site, but you're not using the value, other than checking if it has been specified. That's not how to use a SitePipeBind as that one requires a value in order to be used, E.g. `-Site https://mysiteurl`. Maybe you meant to use a SwitchParameter instead?

Using SitePipeBind would mean people can provide a different url, but you're not resolving it, as you're using ClientContext.Site. The cmdlets will not automatically switch the context based upon the site provided in the SitePipeBind.

Overal: it is very important to make sure that the cmdlets will continue to work as they did before the change, with exactly the same parameters. Otherwise this would be a breaking change and those we only allow in major releases.

**Comment by bhishma** [comment]
Thank you so much for the explanation. SwitchParameter makes a lot more sense as to what I was expecting. Add-PnPEventReceiver has dependency on SiteExtensions in pnpframework.

**Comment by erwinvanhunen** [comment]
Hi,

There is still a problem with the logic in the cmdlet as you propose it. You specify the -Site parameter now as a switch parameter, but it is available for all parameter sets. This means you could theoretically do:

```powershell
Get-PnPEventReceiver -List "mylist" -Site -Identity $identity
```

That does not make much sense... what would be the expected outcome? 

Have a look here for more information about ParameterSets. https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/cmdlet-parameter-sets?view=powershell-7.1

bhishma:
Added another ParameterSet but made ListPipeBind List mandatory on List ParameterSet to resolve ambiguity. Existing functionality still works
Still dependent on https://github.com/pnp/pnpframework/pull/166

KoenZomers:
I'll fix this PR so it works well. Already updated the Add and Get methods. Still need to do the Remove. I will push it out to a new PR once done. I'll therefore close this PR.

Thanks @bhishma for your efforts!

KoenZomers:
Reference to the new PR branch: https://github.com/KoenZomers/pnppowershellcore/tree/PR250

PR #266, opened 2021-02-12
Fix the RSACng cast issue

michael-jensen:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
None that I am aware of - the issue was introduced in my recent [PR 249](https://github.com/pnp/powershell/pull/249)

When running the Disconnect-PnPOnline cmdlet on a connection that was initiated with a AAD App (certificate on the filesystem), the certificate.PrivateKey cannot be cast to RSACryptoServiceProvider as it is currently of type RSACng. Therefore, this PR safely casts to RSACng so the temp cert file can be cleaned up.

@erwinvanhunen sorry I missed this on my original [PR 249](https://github.com/pnp/powershell/pull/249)

### DIFF
--- a/src/Commands/Base/PnPConnectionHelper.cs
+++ b/src/Commands/Base/PnPConnectionHelper.cs
@@ -402,8 +402,11 @@ namespace PnP.PowerShell.Commands.Base
             }
             if (Utilities.OperatingSystem.IsWindows())
             {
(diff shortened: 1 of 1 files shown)


erwinvanhunen:
Thank you!

Issue #273, opened 2021-02-14
Restore-PnPRecycleBinItem: -All invalid parameter


wmadm90:
error message indicates that -All is not a valid parameter, even though it's documented here as an example: https://github.com/pnp/powershell/blob/dev/documentation/Restore-PnPRecycleBinItem.md

unsure if this is an error in documentation or the code?

bhishma:
Looks like the documentation (example 3) needs to be updated.
If you'd like to  restore all items, you can omit  rowlimit.
Just `Restore-PnPRecycleBinItem` will give you what you need.

LeonArmston:
I raised a [PR](https://github.com/pnp/powershell/pull/278) for this to update the documentation and ask what the intended behaviour is regarding using Restore-PnPRecycleBinItem on it's own as currently the -identity parameter is mandatory.

erwinvanhunen:
The documentation has been updated, thank you @LeonArmston ! I will close this issue now.

PR #278, opened 2021-02-15
Updated  Restore-PnPRecycleBinItem documentation

LeonArmston:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Mentioned in https://github.com/pnp/powershell/issues/273#issuecomment-778841403

Updated the documentation for Restore-PnPRecycleBinItem to remove -All parameter which is no longer there.

Although I'm not sure on what the behaviour should be for Restore-PnPRecycleBinItem on it's own? Should it restore all items without the need for a piped input or -Identity parameter? Would you use -RowLimit to limit the amount of items returned and restored? I note -Identity is a required field which means it doesn't allow this behaviour.

### DIFF
--- a/documentation/Restore-PnPRecycleBinItem.md
+++ b/documentation/Restore-PnPRecycleBinItem.md
@@ -39,10 +39,10 @@ Restores all the items in the first and second stage recycle bins to their origi
 
 ### EXAMPLE 3
 ```powershell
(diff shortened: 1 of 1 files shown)


erwinvanhunen:
Thank you @LeonArmston !

PR #284, opened 2021-02-15
Update Get-PnPTenantRecycleBinItem.md

ToddKlindt:
Added some clarifying help around which permissions to use and which command to use depending on which permissions you have. I'll add similar text to the Tenant version.

## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [x] Sample

Additions and clarifications to help

### DIFF
--- a/documentation/Get-PnPTenantRecycleBinItem.md
+++ b/documentation/Get-PnPTenantRecycleBinItem.md
@@ -55,9 +55,4 @@ Accept wildcard characters: False
 ## RELATED LINKS
 
 [Microsoft 365 Patterns and Practices](https://aka.ms/m365pnp)
-Clear-PnPRecycleBinItem
-Clear-PnPTenantRecycleBinItem
-Get-PnPRecycleBinItem
-Move-PnPRecycleBinItem
-Restore-PnPRecycleBinItem
-Restore-PnPTenantRecycleBinItem
+
--- a/documentation/Get-PnPTenantRecycleBinItem.md
+++ b/documentation/Get-PnPTenantRecycleBinItem.md
@@ -24,7 +24,7 @@ Get-PnPTenantRecycleBinItem [-Connection <PnPConnection>] [<CommonParameters>]
 ```
 
 ## DESCRIPTION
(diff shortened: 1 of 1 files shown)


erwinvanhunen:
Thank you Todd! There seems to be something wrong in the PR though. We check for the validity of the documentation, and in the example you pasted a link with no context as such. Also in the related sections the parser expects actual links. If you click on 'details' above where we build the PR you'll see whats wrong.

erwinvanhunen:
Sorry to keep bugging you: but you actually do not have to connect to the tenant admin site for this to work. The cmdlet will do all the things behind the scenes for you. The only requirement is that you are actually a tenant admin. But new Connect-PnPOnline statement is needed. This actually applies to -all- tenant admin cmdlets. They all switch behind the scenes to the tenant admin site automatically, do the work and then switch back to the originally connect to site. E.g.

```powershell
Connect-PnPOnline -Url https://contoso.sharepoint.com # connect to tenant root
Get-PnPList # Show the list in the tenant root
Get-PnPTenantSite # will connect behind the scenes to https://contoso-admin.sharepoint.com and lists all the sites within the tenant
Get-PnPList # Will show again the lists of the the tenant root
```

In scenarios where a company has a vanity url for the tenant they can provide their own url for automatic context switching like this:

```powershell
Connect-PnPOnline -Url https://contoso.sharepoint.com -TenantAdminUrl https://vanityurl.to.tenantadmin.com
```

ToddKlindt:
You're not bugging me, I want the documentation to be correct so it's as easy for folks to use as possible. :)
I understand what you're saying about it switching behind the scenes to do tenant admin work regardless of what URL they connect with. It's a great feature, but don't we want to encourage people to use the tools the most correct way? I feel like it's better advice to say, "Connect to the Admin URL..." than to say, "Connect to whatever site you want, we'll figure it out in the background." 
I'm good either way though. No hurt feelings if you remove that part. :)

erwinvanhunen:
I'm more tempted to say: just connect. Doesn't matter if you connect to the root or tenant admin. Connecting to the tenant admin is a CSOM construct, which has no real 'user' reason but only a technical reason. From a user perspective, if they are a tenant admin, e.g. have the role, then they should be able to execute tenant cmdlets regardless of where they connect to.

PR #553, opened 2021-04-06
Usage clarification for GetDocumentMetadataViaSiteIdUniqueId.ps1

roansmarques:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [x] Sample

## Related Issues? ##
None

Many customers are trying to use this script following a non-interactive PnP Connection, which is leading them to raise support tickets. Without the "interactive" parameter in the connection the result will always be this error: "Submit-PnPSearchQuery : Sequence contains no elements"

### Guidance ###
Added comments to improve the usage experience.


gautamdsheth:
@roansmarques - apologies for the delay here. 
For the issue mentioned, `Submit-PnPSearchQuery : Sequence contains no elements`, have to tried to check with `-EnableQueryRules $False` parameter ? It could be that there are multiple query rules configured.

gautamdsheth:
Sorry , actually this is already clearly mentioned in the readme file of the sample.
Will have to close this PR.

PR #731, opened 2021-05-27
Improve error message in Find-PnPFile when the specified list/folder is not found

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
None

This PR improves the error message displayed by Find-PnPFile cmdlet when passing a list or folder that don't exist.

Before this PR the error message would be "Object reference not set to an instance of an object" when specifying a list that doesn't exist and "File Not Found" when a folder doesn't exist:
image

With this PR the error message for the List case changes to "The specified list was not found":
image

The Folder case still shows the same message because FolderPipeBind.GetFolder() calls web.GetFolderByServerRelativePath(), then Load() the result and finally calls web.Context.ExecuteQueryRetry() which triggers a ServerException. It might be good to make FolderPipeBind.GetFolder() return null if a ServerException happens but that fits better in another PR, even better if a maintainer decides how to handle the case.
For now I left the code handling a null folder object anyway.

Please let me know if I should change anything.

### DIFF
--- a/src/Commands/Files/FindFile.cs
+++ b/src/Commands/Files/FindFile.cs
@@ -1,4 +1,5 @@
-﻿using System.Management.Automation;
+﻿using System;
+using System.Management.Automation;
 using Microsoft.SharePoint.Client;
 
 using PnP.PowerShell.Commands.Base.PipeBinds;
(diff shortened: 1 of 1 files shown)


gautamdsheth:
Thanks @jackpoz  for the PR !
We merged it.
For the folder null exception, would be great if you can submit a separate PR , happy to check that.

PR #896, opened 2021-07-06
When a label for a term is created , it is set as default even if Switch parameter isDefault has not been included

reshmee011:

## Type ##
- [ ] Bug Fix
## Related Issues? ##
When a label for a term is created , it is set as default even if Switch parameter isDefault has not been included

Whenever cmdlet New-PnPTermLabel  is used to create a label, the label is created as default even though -IsDefault is not specified. The change is to allow labels to be created as non-default is the switch parameter -IsDefault is not specified.

The line 66 in file NewTermLabel.cs has been updated so that is switch parameter is not provided, the default is false rathet than true.


gautamdsheth:
Thanks @reshmee011 , it has been merged !
Will be available in tomorrow's nightly.

PR #1093, opened 2021-09-10
Fix null exception when mixing connections with Certificate and without

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
No linked issues

When mixing `Connect-PnPOnline` calls with `ClientSecret` and with `CertificatePath` parameters, a null-exception is currently thrown when trying to access the certificate in the current connect.
This PR adds null checks where required to avoid the exception.

Here's a screenshot from a debugging session in Visual Studio showing that `PnPConnection.Current?.Certificate` is null, so accessing `Thumbprint` triggers the exception:
image

### Guidance ###
* You can delete this section when you are submitting the pull request.* 
* *Please update this PR information accordingly. We use this as part of our release notes in monthly communications.*
* **Please target your PR to Dev branch. If you do not target the Dev branch we will not accept this PR.**


gautamdsheth:
Thank you @jackpoz , have merged the PR !

PR #1166, opened 2021-09-24
Added Connect-PnPOnline command before Url

asadrefai:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [ ] Sample
- [x] Content Update

I might be wrong, but felt adding "Connect-PnPOnline" command before Url to make better understanding. For a beginner it might be confusing having just a Url above the actual command. You guys are better judge :)

### DIFF
--- a/documentation/Get-PnPTenantRecycleBinItem.md
+++ b/documentation/Get-PnPTenantRecycleBinItem.md
@@ -30,7 +30,7 @@ This command will return all the items in the tenant recycle bin for the Office 
 
 ### EXAMPLE 1
 ```powershell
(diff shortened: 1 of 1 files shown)


KoenZomers:
Well noticed @asadrefai that the URL in the sample was odd. We never mention the connect in the examples as there are too many flavors of them, so it would be better just to remove the URL from the sample. I have updated your PR to do so. Thanks for the crisp eye and contribution!

asadrefai:
Thank you @KoenZomers. Pleasure is mine to contribute back to the community :).

PR #1207, opened 2021-10-04
Update New-PnPSite.md

WimVandierendonck:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
The -Lcid parameter description does not contain an explanation about the fact that SharePoint does not support all possible IDs, while is seems to be that way. Copied the relevant text from the '**new-pnptenantsite**' page (https://docs.microsoft.com/en-us/powershell/module/sharepoint-pnp/new-pnptenantsite?view=sharepoint-ps), where it does get mentioned.

Guidance about the SharePoint support for a limited number of IDs for the -Lcid parameter

### DIFF
--- a/documentation/New-PnPSite.md
+++ b/documentation/New-PnPSite.md
@@ -242,7 +242,7 @@ Accept wildcard characters: False
 ```
 
 ### -Lcid
-The language to use for the site.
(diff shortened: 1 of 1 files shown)


KoenZomers:
Good catch @WimVandierendonck ! Thanks for your contribution. I see the Wiki article you're referencing is on our previous repository. I've copied over that article to this repository and updated the link in both cmdlets. Resubmitting this through https://github.com/pnp/powershell/pull/1208 as I don't seem to have permissions to update your PR.

PR #1240, opened 2021-10-12
Allow Sync-PnPSharePointUserProfilesFromAzureActiveDirectorys Users input from valid PnP.Framework.Graph.Model.User lists

reusto:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

Updates Sync-PnPSharePointUserProfilesFromAzureActiveDirectorys User input value to explicitly require a list of PnP.Framework.Graph.Model.User type.

Before this it had to be an array of PSObject and PnP.Framework.Graph.Model.User which therefor only worked with values returned directly by a PowerShell command (Example 1) but not when it's a property of the return (Example 2). The later failed with the following error:

> Unable to cast object of type 'PnP.Framework.Graph.Model.User' to type 'System.Management.Automation.PSObject'.

Now it supports both scenarios.

### Example 1

```powershell
$users = Get-PnPAzureADUser -Filter "jobTitle eq 'IT Administraror'"
Sync-PnPSharePointUserProfilesFromAzureActiveDirectory -UserProfilePropertyMapping @{"CostCenter"="ext_123456"} -Users $users
```

### Example 2

```powershell
$deltaUsers = Get-PnPAzureADUser -Filter "jobTitle eq 'IT Administraror'" -Delta
Sync-PnPSharePointUserProfilesFromAzureActiveDirectory -UserProfilePropertyMapping @{"CostCenter"="ext_123456"} -Users $deltaUsers.Users
```

While debugging this I wondered if the creation of the "aadUsers" object is still needed in this configuration as it is the same input in the end as when getting the users in the command itself when no -Users parameter input is provided. Maybe this step can be saved as well?

Input regarding this is welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added optional `-ScheduledPublishDate` parameter to `Add-PnPPage` and `Set-PnPPage` to allow for scheduling a page to be published.
 - Added `-RemoveScheduledPublish` to `Set-PnPPage` to allow for a page publish schedule to be removed.
 - Added support for off peak SharePoint Syntex content classification and extraction for lists and folders via new `-OffPeak` and `-Folder` parameters for `Request-PnPSyntexClassifyAndExtract`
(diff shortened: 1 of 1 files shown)


reusto:
Ah okay, good to know about the independence from PnP Framework.
Glad I could help.

PR #1314, opened 2021-10-29
Feature - Adding some improvements for IB

gautamdsheth:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
NA

Added necessary parameters to `Set-PnPTenant` and `Set-PnPTenantSite` for Information barrier settings


src/Commands/Admin/SetTenantSite.cs
```diff
@@ -160,6 +160,9 @@ public class SetTenantSite : PnPAdminCmdlet
 
         [Parameter(Mandatory = false, ParameterSetName = ParameterSet_PROPERTIES)]
         public SiteUserInfoVisibilityPolicyValue OverrideBlockUserInfoVisibility { get; set; }
+                
+        [Parameter(Mandatory = false)]
+        public string InformationBarriersMode { get; set; }
```
KoenZomers, line 165: Shouldn't we enum this, if it can only be specific values as you indicate in the documentation proposal above? In that case people can easily tab through the options.

PR #1460, opened 2021-12-17
Support export promoted results to Bookmarks compatible CSV file

wobba:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

Added support to call REST endpoint to fetch promoted results defined via query rules and output them in Bookmark supported CSV format.


gautamdsheth:
Thanks @wobba ! 
Much appreciated :)

PR #1534, opened 2022-01-16
Feature #529 - added cmdlet to retrieve list item permissions

### LINKED ISSUE #529: [FEATURE] Get-PnPListItemPermissions
### Description
I need to know the list item permissions. I know that there are some items with breaked role inheritance but I found no (easy) solution to get the permissions.

### Alternative
Actually I found no solution to get unique role assignments. The following code doesn't work for me.
```
$itemsAll = Get-PnPListItem -List "Site Pages" -Fields "_UIVersionString","EffectiveBasePermissions","EffectiveBasePermissionsForUI","HasUniqueRoleAssignments"
$itemsAll | select { $_.HasUniqueRoleAssignments; }
```

There has been no result.

gautamdsheth:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #529 

Added `Get-PnPListItemPermissions` cmdlet to retrieve list item permissions


KoenZomers:
When I run this, I get an output such as:
image

representing these permissions:

image

Can we not make this a little more useful? The PrincipalId by itself says nothing. We would have to use Get-PnPGroup or Get-PnPUser to find out what's behind the principal. And we don't know in advance which of these two to use.

gautamdsheth:
@KoenZomers - should be more useful now.

image

KoenZomers:
@gautamdsheth I don't believe the permissions property is very useful this way. It was better how it was initially. The PrincipalName property is more useful already. I wonder though if we can make it even better by having a property or some kind of indicator making it clear that its a SharePoint Group or a user? One could check if it starts with i:0#.f, but that feels a bit flaky. Would there be an easy better way you think?

image

gautamdsheth:
You are right, have fixed it now :) 

image

KoenZomers:
Much better! Last nag from my side, do you think it would be possible to also add a flag if the permission is inherited from its parent or is unique to the item? That would also be very valuable. After that I stop nagging :)

Issue #1529, opened 2022-01-14
Get-PnPFile not releasing local file resource when loading file on Mac


gingemonster:
PSVersion 7.2.1
PSEdition Core
OS: macOS Catalina
Pnp.PowerShell 1.9.0

### Issue Summary
Using Get-PnPFile to download a file to the local disk causes file access issues when later used in Add-PnPFileToSiteTemplate throwing error "The process cannot access the file 'filename' because it us being used by another process".

At a guess on MacOS the stream that was used to create the file is not getting closed / recycled.

Working fine on windows, not working on Mac

### Expected behavior 
Should download file and allow adding to site template

### Actual behavior 
Throws file access error on Mac OS

### Steps to reproduce behavior
```
# get temporary files path
$TemplatePath = [System.IO.Path]::GetTempPath()

# connect to sharepoint site
Connect-PnPOnline -Url https://yourdomain/sites/yourexistingsite -Interactive -LaunchBrowser

# download site template
Get-PnPSiteTemplate -Out $TemplatePath + '/mysitetemplate.xml' -IncludeAllClientSidePages -PersistBrandingFiles -PersistPublishingFiles -IncludeNativePublishingFiles -Force

# download a file that exists on you site to the local disk
Get-PnPFile -ServerRelativeUrl '/sites/yourexistingsite/your document library/yourfile.txt' -Path $TemplatePath -FileName yourfile.txt -AsFile

# try and add file to site template
Add-PnPFileToSiteTemplate -Path $TemplatePath + '/mysitetemplate.xml' -Source $TemplatePath + '/yourfile.txt' -Folder '/your document library' -FileLevel Published
```

### What is the version of the Cmdlet module you are running?
Pnp.PowerShell 1.9.0

### Which operating system/environment are you running PnP PowerShell on?
macOS Catalina

veronicageek:
@gingemonster - I'm not very familiar with the `[System.IO.Path]::GetTempPath()` but trying on a Mac it's already giving me some issues. Doing a bit of research, I have found this [closed issue](https://github.com/dotnet/runtime/issues/27771) which is what I get on MacOS Monterey.

And looking at the [Microsoft documentation](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettemppath?view=net-6.0&tabs=windows), I don't see MacOS. But once again, I'm not familiar with it.

Does it still download the file when you run the script? (without the last line `Add-PnPFileToSiteTemplate`......)

[...]

PR #1569, opened 2022-01-27
Fixes Add-PnPFileToSiteTemplate

brannmar:
Fixes an issue where the FileStream tried to overwrite an already open filestream when a file was located in the same directory as the template file itself

## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
#1529

Fixes Add-PnPFileToSiteTemplate

Fixes an issue where the FileStream tried to overwrite an already open filestream when a file was located in the same directory as the template file itself.

Add-PnPFileToSiteTemplate throwing error "The process cannot access the file 'filename' because it us being used by another process".

Also fixed an spelling mistake for the value of const parameterSet_REMOTEFILE.


gautamdsheth:
Thank you @brannmar for the fix !
Much appreciated the investigation and the PR :)

PR #1578, opened 2022-01-28
telemetry as singleton pattern

### LINKED ISSUE #1247: [BUG]Connect-PnPOnline has a memory leak
### Expected behavior 
I should be able to run Connect-PnPOnline thousands of times in a row for different SPO URLs without a System.OutOfMemoryException.

### Actual behavior 
After about 900 to 1,000 times of calling Connect-PnPOnline, I get a System.OutOfMemoryException.

-Create a loop that iterates through 1,000+ valid SPO URLs. (You may be able to cycle through repeat URLs if you don't have 1,000)
-Connect-PnpOnline to each one of the URLs. For me, I am using certificatePath -ClientId -CertificatePassword along with -Tenant and -Url properties.
-Your loop doesn't have to do anything else, but connect. I also tried using Disconnect-PnPOnline afterwards with no luck. Based off of this article (https://docs.microsoft.com/en-us/powershell/module/sharepoint-pnp/disconnect-pnponline?view=sharepoint-ps), that shouldn't be needed anyway.

1.7.0

- [ ] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [X] Azure Functions
- [ ] Other : please specify

gbordier:
## Type ##
- [x ] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ## 

Fixes #1247, 

Implementing a singleton pattern on TelemetryClient to avoid memory leak with repeating connect-pnponline calls

### DIFF
--- a/src/ALC/ApplicationInsights.cs
+++ b/src/ALC/ApplicationInsights.cs
@@ -6,21 +6,44 @@ using System.Linq;
 
 namespace PnP.PowerShell.ALC
 {
+    // implement telemtryclient as a singleton to avoid memory leaks
+    public static class TelemetryClientFactory
+  {
+    private static TelemetryClient _telemetryClient;
+ 
+    public static TelemetryClient GetTelemetryClient()
+    {
+            
+      if (_telemetryClient == null)
+      {
+        
+            TelemetryConfiguration config = TelemetryConfiguration.CreateDefault();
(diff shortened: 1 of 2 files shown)


KoenZomers:
Thanks @gbordier, really useful!

PR #1581, opened 2022-01-31
New cmdlet Get-PnPListDesign

reshmee011:
## Type ##
-  New Feature

New cmdlet Get-PnPListDesign to retrieve List Designs that have been registered on the current tenant. When providing a name with -Identity, it returns all list designs with that name.


KoenZomers:
Thanks @reshmee011, nice one!

Issue #1706, opened 2022-03-18
[BUG] Cannot Connect-PnPOnline when MFA uses PKI/client certificates


mikevasiloff:
### Reporting an Issue
Regardless if `-UseWebLogin` or `-Interactive` is used, when MFA is used **and** requires selection of a PKI/client certificate, the pop-up browser window never shows the certificate selection window.

### Expected behavior 
When connecting to a tenant that requires PKI/client certificate to authenticate, accessing in a normal browser correctly displays a PKI/client certificate selection window and facilitates successful authentication.  This same behavior is what's expected of the PowerShell module.

### Actual behavior 
Using the PowerShell module, although the pop-up browser window does _initially_ seem to work as it loads the authentication page, when it comes to selecting a PKI/client certificate it never actually shows the certificate selection window and the spinning icon just freezes.  You can click on "Sign in with other options" but that just returns you to the screen where you had to click to sign in with a PKI/client certificate.  All you can do then is just close the authentication pop up window which of course fails to connect to the site.
![image](https://user-images.githubusercontent.com/8918397/159006627-a81323bc-3471-4679-bc3f-e709bd3ff390.png)

### Steps to reproduce behavior
`Connect-PnPOnline https://dod365.sharepoint-mil.us -UseWebLogin`

At the sign-in prompt, when asked for an email address enter: _some.user@mail.mil_
When redirected to the sign-in prompt, click on the "Sign in with CAC/PIV" link below the sign-in button.
![image](https://user-images.githubusercontent.com/8918397/159007189-71ddcc1d-1b85-4cc2-8de2-9dc56b9f0b51.png)

What _should_ happen is that a client certificate (smart card) selection window is displayed, but nothing happens.

### What is the version of the Cmdlet module you are running?
1.9.0

### Which operating system/environment are you running PnP PowerShell on?
- [X] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

### Potential fix
What we've found is that if you change the default value for `scriptErrorsSuppressed` to `false` in the [GetWebLoginClientContext function](https://github.com/pnp/powershell/blob/dev/src/Commands/Utilities/BrowserHelper.cs#L27), then the pop-up browser window **is** able to show the PKI/client certificate selection.  But it is unclear why this parameter was set to `true` in the first place.
[...]

PR #1707, opened 2022-03-18
Changed scriptErrorsSuppressed to false

mikevasiloff:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Addresses bug #1706 so that the pop-up browser authentication window can display the PKI/client certificate selection window.

Single change to the src/Commands/Utilities/BrowserHelper.cs file, [line 27](https://github.com/pnp/powershell/blob/dev/src/Commands/Utilities/BrowserHelper.cs#L27) changing the GetWebLoginClientContext function function so that the default value for scriptErrorsSuppressed is `false`.


KoenZomers:
Thanks for your contribution and detailed repro steps @mikevasiloff . I believe to remember from firsthand that the errors are being suppressed as at least previously it could happen a lot that the old Internet Explorer container would not be able to handle parts of the login pages very well and would throw several JavaScript error dialog boxes when trying to log on. Not sure if that's still the case today. When testing with your PR branch against my normal tenant, I did not get any errors anymore, so I guess we're good with this change if that helps in your scenario.

One thing I'm wondering though, you're mentioning the issue you're facing applies to both UseWebLogin and Interactive. These are both completely different implementations. Only the first would be affected by the change you propose, so I don't see how the Interactive login would be fixed by this. Have you tested this and are you sure this is the case?

mikevasiloff:
@KoenZomers, you're very welcome.  Ok that makes sense, thanks for the background info.

As for `-Interactive`, I'll have to double check as I thought I tested both.  For the National Clouds we have to register our own app in Azure AD for "PnP PowerShell" and then specify the `-ClientId`, `-Tenant`, and `-AzureEnvironment` when making a connection which still opens the same user PKI/client certificate selection window.

KoenZomers:
Please check indeed if interactive works okay as well or needs additional fixing. Weblogin is an old method which only provides access to the SharePoint API and not any of the countless other APIs behind many of the cmdlets so best would be if interactive or devicelogin will work for your scenario.

mikevasiloff:
@KoenZomers, I completely understand about the limitations of the `-UseWebLogin` option, it's just that in the Sovereign Clouds it can be hard to convince customers to register PnP to enable the `-Interactive` method.  After looking at this again you are right that even with the change I proposed that helps the `-UseWebLogin` case, it didn't affect the `-Interactive` case.  When using `-Interactive` it exhibited the same behavior as documented in bug #1706 where the authentication pop-up window just "freezes" rather than providing the ability to select the PKI/client certificate.  Unfortunately, I'm not familiar enough with the code to know which function relates to the `-Interactive` parameter.

mikevasiloff:
Actually, looking at your change to the PnPConnection.cs file I noticed `CreateWithInteractiveLogin` with a call to the `PnP.Framework.AuthenticationManager.CreateWithInteractiveLogin` function.  So it looks like this is using a PnP Framework method.  This then lead me to the [PnP.Framework/AuthenticationManager.cs](https://github.com/pnp/pnpframework/blob/dev/src/lib/PnP.Framework/AuthenticationManager.cs) file where I see `WithCustomWebUi` being called, but I can't seem to find the source for that.  If this is ultimately coming from the PnP Framework instead, does that mean I need to open a bug in that repo for the `-Interactive` part of this issue?

KoenZomers:
Would have to dive into the code to be 100% sure but I believe it's actually MSAL that is being utilized for interactive and devicelogin. Have you tried your success with devicelogin? This allows you to use any browser thus is not dependant on possible limitations by any framework.

KoenZomers:
As we're going to do a new release of PnP PowerShell any time now, I will proceed with merging this PR so it will be included in this next release. If any further work needs doing to accommodate other authentication methods for making use of certificates, please file a new PR and we can take it from there.

PR #1761, opened 2022-04-11
Build-Debug.ps1: improve error message when PowerShell is too old

fowl2:
Without this the error is
`Unable to find type [System.Management.Automation.SemanticVersion].`


gautamdsheth:
@fowl2 - thank you for this contribution !

Have merged it.

PR #1763, opened 2022-04-11
Principal CmdLets: add OutputType attribute

fowl2:
Adds `[OutputType]` attributes to the Principal related cmdlets where it makes sense.

Makes autocomplete work eg. type `(Get-PnPUser).` and press tab and it'll show a nice menu if you've got it configured like that*.

image

*ie. `Set-PSReadlineKeyHandler -Key Tab -Function MenuComplete`


KoenZomers:
Really neat one @fowl2, thanks for this contribution!

fowl2:
Would this kind of improvement be welcome across the codebase? Would you prefer one PR per folder like this?

KoenZomers:
Sure thing this would be welcome! As it will likely touch a lot of files, chunking them into several PRs would indeed be nice. Whether by folder or just one for every, say 20, cmdlets, would be nice. Only thing I changed to your PR was to put models into the models folder and not create an additional class inside the cmdlet class. This to align with how other coding has been done already within the module. If you could stick to that for any new models you may be creating, that would be nice.

PR #1781, opened 2022-04-20
Lists cmdlets: Add missing OutputType attributes for better autocompletion

fowl2:
Adds the `OutputType` attribute for improved autocompletion on the cmdlets in the Lists folder.

There's a few in here that seem to be missing outputs (created comment, updated list, recycle bin item, etc.) I'll do those in separate PRs as they might need some discussion.

Follow up to [PR #1763](https://github.com/pnp/powershell/pull/1763#issuecomment-1099695169)


gautamdsheth:
Thanks @fowl2 😊🙏

PR #1808, opened 2022-04-26
Consistency when retrieving Site Group by name (Closes #1721)

### LINKED ISSUE #1721: Set-PnPGroupPermissions Case Sensitivity Inconsistency
Set-PnPGroupPermissions -Ideneity "Group Name" <- "Group Name" **is** expected to be case sensitive. However,

Get-PnPGroup -Identity "Group Name" <- "Group Name" **is not** expected to be case sensitive.

The issue comes about when you use Get-PnPGroup to check to see if a group exists before you use Set-PnPGroupPermissions to set its permissions. Since the later is case sensitive, it could fail even though logically the group exists.

martinlingstuyl:
Consistency when retrieving Site Group by name.

## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #1721

Added some consistency between `Get-PnPGroup`, `Get-PnPGroupPermissions` and `Set-PnPGroupPermissions` commandlets. When using the name parameter, the group is queried through the PnP Core SDK, which is case sensitive. Also added consistent `null` checks with `PSArgumentExceptions` when group was not found.

### DIFF
--- a/documentation/Get-PnPGroup.md
+++ b/documentation/Get-PnPGroup.md
@@ -59,7 +59,7 @@ Returns all SharePoint groups in the current site
 Get-PnPGroup -Identity 'My Site Users'
 ```
 
-This will return the group called 'My Site Users' if available in the current site
+This will return the group called 'My Site Users' if available in the current site. The name is case sensitive, so a group called 'My site users' would not be found.
 
 ### EXAMPLE 3
 ```powershell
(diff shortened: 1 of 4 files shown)


gautamdsheth:
Awesome , thanks @martinlingstuyl for this update !
Much appreciated :) 

Didn't think myself that a case sensitivity issue would go so far, earlier I just thought that maybe a `ToLower` or something would be enough 😂😊

martinlingstuyl:
It's new for me as well. Another solution would have been to load the titles of all sitegroups in memory, and then compare them clientside with Case Sensitivity disabled... but well, that just feels wrong, doesn't it? 😂 anyway, if there happens to be a better approach, we can change it again!

PR #1859, opened 2022-05-12
Update New-PnPSite.md

jagopauwels:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
No

Change to the documentation of New-PnPSite: When the parameter -IsPublic is not specified the default value is false, so the group will be Private. The current documentation incorrectly says it will be Public.

### DIFF
--- a/documentation/New-PnPSite.md
+++ b/documentation/New-PnPSite.md
@@ -226,7 +226,7 @@ Accept wildcard characters: False
 ```
 
 ### -IsPublic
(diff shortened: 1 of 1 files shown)


gautamdsheth:
Thank you @jagopauwels , have merged the PR ! 
Is an important contribution indeed , much appreciate :) 
The updated docs will be available from tomorrow !

PR #1888, opened 2022-05-21
Fix: Get-PnPField now returns specific field type

gautamdsheth:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
NA

Making the behavior consistent with the other parameters, `Get-PnPField`  now returns specific field type instead of the generic field when even List parameter is specified.

When using `Add-PnPField`, it was not working correctly for taxonomy types, it is now fixed.


KoenZomers:
@gautamdsheth I understand why you're doing this, but I think this will create a non backwards compatible change, right?

gautamdsheth:
Yeah unfortunately ☹

KoenZomers:
@gautamdsheth I don't think it will be wise to break backwards compatibility with this. Would it be an idea to add a parameter -ReturnTyped which then would return the fields as typed fields and without this flag it would keep it as it is today?

gautamdsheth:
Yeah , I can add that, no issues. Its just that it is inconsistent. If you dont specify the `-List` parameter, then it returns the typedValue, but if specify it returns a generic one. 

Totally agree, it might break backward compatibility, but adding is the additional param, should we really do that ??

KoenZomers:
I don't really know another option. It's either that or do a non backwards compatible change.

gautamdsheth:
@KoenZomers , changed the implementation :)

PR #1893, opened 2022-05-23
Updated example of the query parameter

bjdekker:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [x] Sample

The two examples of the Add-PnPDataRowsToSiteTemplate cmdlet contained a value for the Query parameter that caused confusion. At line 98 in the cmdlet, the query is merged with the value of the fields list inside a <View></View> element:
`query.ViewXml = string.Format("<View>{0}{1}</View>", Query, viewFieldsStringBuilder);`

So in order for this to work, the Query parameter should contain a <Query> element. I removed the Query parameter from the first example since it does not do anything, it's only confusing and added a more meaningful and working example to the second example.


gautamdsheth:
Thank @bjdekker for this !

Makes sense and is much more clearer now :)

Issue #1580, opened 2022-01-31
Distributing the Module as a Docker Image


shurick81:
**Is your feature request related to a problem? Please describe.**

As a developer working with a multitude of projects where PnP.Powershell is used, I would like to be able to quickly run the module without the need of actually installing the module in the host environment so that I have isolated and repeatable experience of using the module for multiple different project.

**Describe the solution you'd like**

Every time when a new version of the PnP.PowerShell module is released, we could

1. build a new a new version of the Docker image with the module installed. 

I tried the following Dockerfile locally and it works quite fine for me so far:

```docker
FROM mcr.microsoft.com/powershell:7.2.1-alpine-3.14-20211215

RUN pwsh -c "Install-Module -Name PnP.PowerShell -RequiredVersion 1.9.0 -Force"
```

2. do some testing of the new module version working inside a container
3. publish the image to https://hub.docker.com/u/m365pnp

**Describe alternatives you've considered**

Alternatively we could just come up with a small guide/demo/webinar that shows how to build and use a local image, since there is no published one.

**Additional context**

We wanted to consider this improvement when the module GA is close enough, but then this issue was forgotten I believe:
https://github.com/pnp/powershell/issues/11

KoenZomers:
@shurick81 I have no such experience myself to be able to set this up. Would you be able to do this?

shurick81:
@KoenZomers yes, I would love to do it. First of all we could discuss some more specific use cases here in this thread and then I could come up with a PR.

KoenZomers:
Nice. I think you mention one yourself already in your start post. I could envision having a docker image per version perhaps to test things in a specific version? Not a docker guy here so not sure if that makes sense.

shurick81:
I suggest that if we go with docker image, we need to extend the use cases.

1. to quickly run the module without the need of actually installing the module in the host environment so that I have isolated and repeatable experience of using the module for multiple different projects
2. to access local host directories through using volumes on a docker container
3. to use image for running jobs in such pipeline engines like Azure DevOps (maybe in GitLab also)

What do you think?

shurick81:
Here's a preliminary result:
1. nightly releases example: https://hub.docker.com/repository/docker/asapozhkov/pnp-powershell
2. Build and publish nightly releases: https://github.com/alexsapozhkov/powershell/tree/dev/docker
3. Build and publish via Github Actions: https://github.com/alexsapozhkov/powershell/actions/workflows/nightlyrelease.yml

[...]

PR #1895, opened 2022-05-23
Parametrized Docker organization

shurick81:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes issue in the feature #1580

Using `DOCKER_ORG` for locating docker image repository in Docker Hub


shurick81:
Hi @gautamdsheth, here's the fix for the organization name so that the image is published not into the account but into a parametrized organization.

PR #1900, opened 2022-05-25
Add-PnPDataRowsToSiteTemplate export dates as UTC string

bjdekker:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
N/A

The Add-PnPDataRowsToSiteTemplate cmdlet exports DateTime values as-is. That means that there is no indication that the date is in UTC Time zone. Once you apply the template, the dates are read as local times, which can be a few hours off. This change formats the exported dates as UTC string, so they are not converted to a local datetime when imported.


gautamdsheth:
Thank @bjdekker for this improvement , makes total sense !
Have merged it and will be available in tomorrow's nightly

PR #1902, opened 2022-05-25
Fix #1554 - issue with onedrive site storage quota

### LINKED ISSUE #1554: [BUG] Get-PnPUserOneDriveQuota reporting license maximum quota instead of actual quota
### Reporting an Issue
`Get-PnPUserOneDriveQuota` returns a value that is not equal to actual applied quota. Problem still occurs after using `Set-PnPUserOneDriveQuota`.

### Expected behavior 
The command should return the same quota as the one applied, displayed in administration center or returned by `Get-SPOSite` command. 

### Actual behavior 
The value returned seems to be the maximum quota allowed by the user's license. 

Here's an example getting one user quota and the actual quota applied to underlying SharePoint site collection. 

image

```
$user = "someone@tenantdomain"
$TenantSPOAdminUrl = "https://tenant-admin.sharepoint.com"
$credentials = Get-Credential -Message "SPO admin account"
Connect-SPOService -Url $TenantSPOAdminUrl -Credential $credentials
Connect-PnPOnline -Url $TenantSPOAdminUrl -Credentials $credentials
(Get-PnPUserOneDriveQuota -Account $user | Select-Object -ExpandProperty Value)/(1024*1024)
Get-SPOSite -Identity ((Get-PnPUserProfileProperty -Account $user).personalurl -replace "/$","") | Select-Object -ExpandProperty StorageQuota
```

Using 1.9.0 version of the module

- [x] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

gautamdsheth:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #1554 

Fix the one drive storage quota issue. 
We need to get the storage quota level of the personal site itself and not the max quota.

@KoenZomers - need your help here :) 

1) Can we optimize it more to retrieve the site ? 
2) I had to add 1024 * 1024 , not sure why, but only after that it worked. Seems a bit flaky , need something more robust, any ideas ?


KoenZomers:
Yikes, fetching ALL OneDrive sites in order to do this? There must hopefully be a more efficient way of achieving this. How can this be reproduced anyway? I couldn't repro it before. Is the catch here to set it to something less than the tenant configured maximum?

gautamdsheth:
Yeah, the filter with URL didnt work, so had to fetch all sites :( 

To repro this, you can just do:

```
Set-PnPUserOneDriveQuota -Account 'user@domain.com' -Quota 5368709120 -QuotaWarning 4831838208

```

and then use the above code.

gautamdsheth:
@KoenZomers  - managed to figure out an efficient way to filter the site for a specific user. Looks good to me now :)

PR #1919, opened 2022-05-31
Various fixes when using specific Connections

### LINKED ISSUE #1857: [BUG] Disconnect-PnPOnline -Connection $connection1 closes all connections
The ability to have multiple connection objects and use and disconnect them at will

### Expected behavior 
Running Disconnect-PnPOnline -Connection $connection1 disconnects only connection in $connection1

### Actual behavior 
Running Disconnect-PnPOnline -Connection $connection1 disconnects all connections

image

Latest

- [x ] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [x ] Azure Functions
- [ ] Other : please specify

KoenZomers:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #1857 

Lots of cleanup and refactoring around using a specific connection retrieved by `Connect-PnPOnline -ReturnConnection` or `Get-PnPConnection`. Things were getting completely mixed up. It should now completely be separated from the current context which even would allow having multiple connections at once, each with a different authentication method or user.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed `Invoke-PnPSPRestMethod` invalid parsing for SharePoint number columns. [#1877](https://github.com/pnp/powershell/pull/1879)
 - Fix issue with `Add/Set-PnPListItem` not throwing correct exception for invalid taxonomy values. [#1870](https://github.com/pnp/powershell/pull/1870)
 - Fixed `Sync-PnPSharePointUserProfilesFromAzureActiveDirectory` throwing an "Object reference not set to an instance of an object" exception when providing an empty users collection or incorrect user mapping [#1896](https://github.com/pnp/powershell/pull/1896)
+- Fixed `Connect-PnPOnline -ReturnConnection` also setting the current connection instead of just the returned connection
+- Fixed `Disconnect-PnPOnline -Connection` also disconnecting other connections next to the provided connection
 
 ### Removed
 - Removed `Get-PnPAvailableClientSideComponents`. Use `Get-PnPPageComponent -Page -ListAvailable` instead.  [#1833](https://github.com/pnp/powershell/pull/1833)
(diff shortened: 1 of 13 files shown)


heinrich-ulbricht:
I so hope that this is the solution for https://github.com/pnp/powershell/issues/1949

@KoenZomers Is this already going to the prerelease builds?

KoenZomers:
@heinrich-ulbricht, this is already in the latest nightly builds, yes. I did find a regression with this PR that prevents one from using disconnect-pnponline and then connect-pnponline again after that. I know where the issue lies, just need to discuss within the team what makes most sense in how to fix it.

heinrich-ulbricht:
@KoenZomers Ok I'll have to wait for that because I use Disconnect-PnPOnline followed by Connect-PnPOnline (i.e. connect for all connections, I ran into https://github.com/pnp/powershell/issues/1857 as well).

heinrich-ulbricht:
@KoenZomers I would totally expect that I would have to authenticate again. Like the first time. Disconnect is disconnect.

KoenZomers:
Thanks for your extensive input. I agree that it would be most intuitive to work this way. Problem we're facing is that it hasn't worked this way and even was deliberately coded not to work this way, so to do automatically authenticate when connecting to the same tenant again. Thereby changing this would/could be seen as a non-backwards compatible change, which is something we really try to stay away from with all the scripts people have running unattended. Tricky one. We'll give it some more thoughts what's wise here.

Nathaire:
I think this PR create a breaking change in the flow. We had a script using `$myconnec = Connect-PnPOnline -ReturnConnection` and then later `Disconnect-PnPOnline -Connection $myconnec`

Before we get a warning, but now it throws an exception
image

Should breaking change like this not only be integrated in major version change, not minor one ?

KoenZomers:
@Nathaire This is an unintended regression. Will look into fixing this.

Nathaire:
@KoenZomers thanks a lot

KoenZomers:
The disconnect situation will be fixed by this PR: https://github.com/pnp/powershell/pull/2093

PR #1922, opened 2022-05-31
Maintain orderhint for `Get-PnPPlannerBucket`

### LINKED ISSUE #1918: [BUG]/[Question] - `Get-PnPPlannerBucket` bucket ordering is  lost
In my perspective it's a bug, but it's also a question.
I'm retrieving Planner buckets via `Get-PnPPlannerBucket`. I noticed that the ordering is not the same as in the UI.
This is strange since Graph does maintain the order of the buckets when I retrieve them via Graph explorer.

When I looked at the code I noticed that after the buckets are fetched, a custom sorting is done on the orderHint (line 405).

https://github.com/pnp/powershell/blob/0bb321e5679369eba510c183039de72f0fa95e21/src/Commands/Utilities/PlannerUtility.cs#L402-L406

Since the orderHint is a weird value ([docs](https://docs.microsoft.com/en-us/graph/api/resources/planner-order-hint-format?view=graph-rest-1.0)) sorting it alphabetically is not the way to do it. The original sorting of the buckets are gone and I don't know anymore which bucket is first and which is last.

This is what I have in Planner
image

This is what I get when I use the Graph call:
image

This is what is returned via PnP.PowerShell
image

As you can see, Graph maintains the order (in reverse order). But PowerShell kind of shuffles them.

My question is fairly simple. Can I remove the custom sorting so the Graph sorting is maintained? 😃 

### Expected behavior 
I'd like the buckets to maintain their order.

### Actual behavior 
Buckets are sorted alphabetically on oderHint and thus sort of shuffled.

1. Create a plan with 5 buckets.
2. Execute cmdlet `Get-PnPPlannerBucket`

1.10.31-debug

- [x] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

milanholemans:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #1918

Removed custom sorting so the Graph sorting on `orderHint` property is maintained.


KoenZomers:
Thanks @milanholemans !

PR #1934, opened 2022-06-03
Adds AllowTenantMoveWithDataMigration parameter for `PlannerConfiguration` cmdlets

milanholemans:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
None

Noticed there was a missing property in the `Get-PnPPlannerConfiguration` and `Set-PnPPlannerConfiguration` cmdlets called `AllowTenantMoveWithDataMigration`.
This PR adds the property for both commands.

I didn't find any documentation at all, so I'm not 100% sure what the property does. Hence the rather vague description in the docs. Maybe you guys are able to set a better description?


KoenZomers:
Thanks! I believe I initially left it out because it was either vague or an internal property only, but let's add it anyway.

milanholemans:
> Thanks! I believe I initially left it out because it was either vague or an internal property only, but let's add it anyway.

Agree it's a vague property, but you're able to update the value of it so I also think it's better to include it.

PR #1935, opened 2022-06-05
Introduce plan ID parameter for `Get-PnPPlannerPlan`

milanholemans:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
None

Introduced a new parameter `-Id` to retrieve a plan only by ID without a group ID.

This makes it also more consistent with other Planner cmdlets.


KoenZomers:
Thanks!

PR #1960, opened 2022-06-10
SetList: receive list from pipeline

fowl2:
Adds the ability to do

    Get-PnPList | ? -not EnableVersioning | Set-PnPList -EnableVersioning:$true

instead of the more clumsy:

    Get-PnPList | ? -not EnableVersioning | % { Set-PnPList -Identity $_ -EnableVersioning:$true; }

Also

    Set-PnPList "MyList" -EnableVersioning:$true

Instead of

    Set-PnPList -Identity "MyList" -EnableVersioning:$true

A small improvement, but adds to the overall consistency.


KoenZomers:
Thanks @fowl2 !

PR #2009, opened 2022-06-18
Update Set-PnPWeb.md

PowershellScripts:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [x] Sample

## Related Issues? ##
None

Example, and missing parameter description.

**Question**: Would it perhaps make sense to remove the Header parameters (SiteLogoUrl, HeaderLayout, HeaderEmphasis, HideTitleInHeader) from **Set-PnPWeb** cmdlet since they are already available in **Set-PnPWebHeader** cmdlet and even a bit extended there?

### DIFF
--- a/documentation/Set-PnPWeb.md
+++ b/documentation/Set-PnPWeb.md
@@ -43,10 +43,17 @@ Hides the quick launch from being shown in the current web
 
 ### EXAMPLE 3
 ```powershell
+Set-PnPWeb -HeaderEmphasis Strong -HeaderLayout Compact
+```
+Sets the header style in the current web.
+
+### EXAMPLE 4
+```powershell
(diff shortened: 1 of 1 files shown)


KoenZomers:
Fair point, although I'm a little bit hesitant. We cannot just remove it as existing scripts might rely on it, but we could mark it as obsolete and then remove it half a year down the road or so. If I look at the implementation of SiteLogoUrl, I see that `Set-PnPWeb` uses CSOM to set it while `Set-PnPWebHeader` uses a direct SPO API call to set it. They might eventually do the same, but it's a bit hard to say for sure for all possible scenarios (i.e. group backed, standalone, comms, teams, etc). I'm tempted therefore to play safe and keep them both.

@gautamdsheth  What are your thoughts on the raised question?

PR #2036, opened 2022-06-26
Changes around Classification and Sensitivity Labels

KoenZomers:
## Type ##
- [X] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
N/A

Went through the entire situation around Site Classification and Sensitivity Labels and rewrote it.


gautamdsheth:
@KoenZomers  - Some of these cmdlets are missing `[RequiredMinimalApiPermissions("InformationProtectionPolicy.Read.All")]`, is it by design ??

KoenZomers:
@gautamdsheth Yes. Mainly because of how this has been developed within PnP. As we may execute these cmdlets using a delegate and an app only token, it would require different permissions for either of them. If we specify that attribute, it will error out if we use the other one since it will be missing from the scp in our access token. We would need something where we can specify if it works for delegate tokens and app only tokens. Fairly easy to implement and I believe it was like this originally but at one point Erwin decided to simplify the permissions leaving just this.

PR #2038, opened 2022-06-27
Added Microsoft 365 Group Endpoint cmdlets

KoenZomers:
## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
N/A

- Added `Get-Microsoft365GroupYammerCommunity` cmdlet to retrieve details on the Yammer Community connected to a Microsoft 365 Group
- Added `Get-Microsoft365GroupTeam` cmdlet to retrieve details on the Microsoft Teams team connected to a Microsoft 365 Group
- Added `Get-Microsoft365GroupEndpoints` cmdlet to retrieve details on all endpoints connected to a Microsoft 365 Group


gautamdsheth:
@KoenZomers  - can we please use `Microsoft365GroupPipeBind` here instead of GUID ? Would make it much easier to validate and make it consistent with other M365 Group related cmdlets.
Also, we can then just throw the error if the Group is not found like this, instead of warning:

```
throw new PSInvalidOperationException("Group not found");
```

KoenZomers:
@gautamdsheth Its already using the pipebind?

gautamdsheth:
Sorry, my bad.

About this condition:
```
if(groupId == Guid.Empty)
                {
                    WriteWarning("Current site is not backed by a Microsoft 365 Group");
                }
```


Can we just throw the error for this ? Because, I am guessing the Graph call will also fail for this

KoenZomers:
Good point, well spotted, I'll update it!

KoenZomers:
@gautamdsheth Ready for final review.

PR #2044, opened 2022-06-28
Changes to `Get-PnPSiteCollectionAppCatalogs`

### LINKED ISSUE #2041: [BUG] Get-PnPSiteCollectionAppCatalogs showing old site urls after the site is renamed and url changed
### Notice
Many bugs reported are actually related to the PnP Framework which is used behind the scenes. Consider carefully where to report an issue:

1. **Are you using ```Invoke-PnPSiteTemplate``` or ```Get-PnPSiteTemplate```**? The issue is most likely related to the Provisioning Engine. The Provisioning engine is _not_ located in the PowerShell repo. Please report the issue here: https://github.com/pnp/pnpframework/issues.
No
1. **Is the issue related to the cmdlet itself, its parameters, the syntax, or do you suspect it is the code of the cmdlet that is causing the issue?** Then please continue reporting the issue in this repo.
Code
1. **If you think that the functionality might be related to the underlying libraries that the cmdlet is calling** (We realize that might be difficult to determine), please first double check the code of the cmdlet, which can be found here: https://github.com/pnp/powershell/tree/master/src/Commands. If related to the cmdlet, continue reporting the issue here, otherwise report the issue at https://github.com/pnp/pnpframework/issues
No

Please confirm what it is that your reporting

I am running Get-PnPSiteCollectionAppCatalogs and I realized that if a site url was updated for site collection app catalog its not updated in the output of the command.

Get-PnPSiteCollectionAppCatalogs

### Expected behavior 
Please describe what output you expect to see from the PnP PowerShell Cmdlets

The output will include updated url for the sitecollectionappcatalog instead of the original url

### Actual behavior 
Please describe what you see instead. Please provide samples of output or screenshots.

The output is including the old sitecollectionappcatalog

Please include complete script or code samples in-line or linked from [gists](https://gist.github.com/)

$data = Get-PnPSiteCollectionAppCatalogs

(you can retrieve this by executing ```Get-Module -Name "PnP.PowerShell" -ListAvailable```)

1.10.0

- Windows

KoenZomers:
## Type ##
- [X] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #2041 

- Renamed `Get-PnPSiteCollectionAppCatalogs` to `Get-PnPSitecollectionAppCatalog` to follow our naming convention
- Changed so that `Get-PnPSitecollectionAppCatalog` now returns our own model
- Every returned site will be validated if it still exists and about the actual current URL
- Added `-ExcludeDeletedSites` optional parameter to `Get-PnPSiteCollectionAppCatalogs` which allows for site collections with App Catalogs that are in the recycle bin to be exluded from the results
- Added `-CurrentSite` optional parameter to `Get-PnPSiteCollectionAppCatalogs` which allows for checking if the currently connected to site has a site collection App Catalogs provisioned on it


gautamdsheth:
Hey @KoenZomers , lets not introduce breaking changes here.

Lets just return whatever CSOM is returning instead of our own object.  

Consider this scenario:

```
$appCatalogSites = Get-PnPSiteCollectionAppCatalog
$appCatalogSites[0]
```

Now, if `$appCatalogSites[0]` has some property, like `$appCatalogSites[0].Title` that is being used by users, it will be broken.

Same goes for the page, lets keep the URL as it was before. 
People do keep that as bookmarks and suddenly they will find that its not working.
The alias is good , once we remove that Alias, we will change the page URL as well

PR #2084, opened 2022-07-04
Update Set-PnPPlannerConfiguration.md

PowershellScripts:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [x] Sample

## Related Issues? ##
N/A

Example.

## Question ##
@gautamdsheth  @KoenZomers 
What does it mean when CommonParameters are supported? 
Unlike Set-PnPSite, where -Debug actually enters me into Debug mode, and -ErrorAction does what it's supposed to do; here the Common Parameters are there, available, but they do not seem to do much. Same with e.g. Set-PnPPlannerPlan. Should the CommonParameters be documented in .md file or not?

### DIFF
--- a/documentation/Set-PnPPlannerConfiguration.md
+++ b/documentation/Set-PnPPlannerConfiguration.md
@@ -13,7 +13,7 @@ schema: 2.0.0
 
 * Azure: tasks.office.com
 
-Allows the Microsoft Planner configuration of the tenant to be set
+Allows the Microsoft Planner configuration of the tenant to be set.
 
 ## SYNTAX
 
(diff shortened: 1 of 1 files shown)


KoenZomers:
Good point. These have bothered me for a while already too. Let me discuss within the team what the original mindset was behind it.

PowershellScripts:
@KoenZomers Did you have the time to discuss this one?

KoenZomers:
Yes, this one too! :) We are in favor of no longer mentioning the commonparameters but will not actively remove them everywhere for now. If a cmdlet i.e. implements verbose logging or whatif, we will explicitly mention these as being options in the documentation for that cmdlet. So for you if you happen to be updating documentation and still see it listed, feel free to remove it.

PR #2117, opened 2022-07-08
Update Set-PnPBrowserIdleSignout.md

PowershellScripts:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [ ] Sample
- [x] Formatting

## Related Issues? ##
N/A

In the source -WarnAfter and -SignoutAfter parameters are non-Mandatory. I'm guessing it's because for 
```powershell
 Set-PnPBrowserIdleSignout -Enabled $false
```
they are not needed and there is only one parameter set.

In the Documentation they are Required. I left them as such because 
```powershell
Set-PnPBrowserIdleSignout -Enabled $true -SignoutAfter 0.11:00:00
```
doesn't work without the other parameter and vice-versa.

Feel free to amend with a commit if you feel it should be otherwise.

### DIFF
--- a/documentation/Set-PnPBrowserIdleSignout.md
+++ b/documentation/Set-PnPBrowserIdleSignout.md
@@ -14,18 +14,24 @@ Sets the current configuration values for Idle session sign-out policy.
 
 ## SYNTAX
 
+### Enable
 ```powershell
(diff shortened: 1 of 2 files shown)


KoenZomers:
@PowershellScripts I've added a paramset to better deal with the enabling and disabling. Thanks for noticing this!

PR #2122, opened 2022-07-09
Set-PnPField: add parameters for properties

fowl2:
This PR adds parameters for each of the settable properties across all the different field types in `Set-PnPField`. 

My goals were:
1. Autocomplete shouldn't suggest invalid combinations
2. Shouldn't have to add any additional parameters just to satisfy point 1. 

I think I've achieved it. It took quite a few attempts and for I a while I thought it wouldn't be possible or require an unwieldy number of combinations - but it seems to works

Using dynamic parameters I think there's about 500 parameter+parametersets combos in total to make it all work! Not too bad considering the number of properties.
```
> (Get-Command set-pnpfield).ParameterSets | select Name, { ($_.Parameters |? IsMandatory).Name }, {$_.Parameters.Count}

Name                                                    ($_.Parameters |? IsMandatory).Name  $_.Parameters.Count
----                                                   ------------------------------------- -------------------
Field                                                  Identity                                               46
FieldCalculated                                        Identity                                               51
FieldChoice                                            Identity                                               49
FieldComputed                                          Identity                                               47
FieldCurrency                                          Identity                                               54
FieldDateTime                                          Identity                                               51
FieldGeolocation                                       {Identity, FieldGeolocation}                           46
FieldGuid                                              {Identity, FieldGuid}                                  46
FieldLocation                                          {Identity, FieldLocation}                              53
FieldLookup                                            Identity                                               54
FieldMultiChoice                                       Identity                                               48
FieldMultiLineText                                     {Identity, FieldMultiLineText}                         53
FieldNumber                                            Identity                                               53
FieldRatingScale                                       Identity                                               54
FieldText                                              Identity                                               47
FieldUrl                                               Identity                                               47
FieldUser                                              Identity                                               59
FieldCalculated_FieldCurrency                          Identity                                               18
FieldCalculated_FieldDateTime                          Identity                                               18
FieldCalculated_FieldCurrency_FieldNumber              Identity                                               18
FieldChoice_FieldMultiChoice_FieldRatingScale          Identity                                               19
FieldCurrency_FieldNumber                              Identity                                               23
FieldDateTime_FieldUrl                                 Identity                                               18
FieldLocation_FieldMultiLineText                       Identity                                               23
FieldLocation_FieldLookup_FieldMultiLineText_FieldUser Identity                                               18
FieldLookup_FieldUser                                  Identity                                               24
```

The two main ways of using this cmdlet I've tried to polish are:
1. Specify the type of field, using one of the new switch parameters added for either field type eg. `-FieldChoice`, and then using the completion which will only suggest the applicable properties.
2. Changing fields without caring about their types - eg. Updating the `-Required` property for all fields. No `-Field*` type parameter should be required in this case. This should also work for properties that are shared by different types, even in different hierarchies.  Eg. there are ~18 shared by `FieldCalculated` and `FieldDateTime`.

The `Values` hashtable parameter remains for the flexibility and compatibility. 

Also added returning the updated `Field` because it makes testing so much easier, but could take that out if desired. 

edit: this was draft for a while but it's probably ready.


KoenZomers:
@fowl2 Can you elaborate on how this is supposed to work or how I can test it? When I build this PR branch, it doesn't show me any of the dynamic suggestions it seems.

image

fowl2:
They suggestions are for full parameters, not part of `-Values`.

Try:
* adding `-FieldChoices` and seeing how `-MaximumValue` is not suggested, but `-Choices` is.
* adding `-MaximumValue` and seeing how `-Choices` is not suggested, but `-MinimumValue` is.

KoenZomers:
For me this doesn't work. When I try the second same you give, it does resolve to Choices when specifying MaximumValue

image

The first one seems to work:

image

It does not resolve MaximumValue

fowl2:
This appears to be PowerShell behavior - if there is only one possible parameter, it will be suggested, even if it is invalid. If you typed `Set-PnPField -Identity Something -FieldChoice -MaximumV` it would also complete, but not suggest. 

With the `Set-PSReadLineKeyHandler -Key Tab -Function MenuComplete` set it is more clear(ly not suggested):

image

It will not actually allow the cmdlet run like this:

image

KoenZomers:
I appreciate all the hard work you've put into this @fowl2, but it never really became clear to us in the team what the true added value was and if it works well. I don't feel it is useful to keep this open any longer than the 1+ year it already has. I'm sorry but I'm going to reject the PR as I find this cmdlet too trivial to pass through a change that I don't fully understand or can test.

PR #2126, opened 2022-07-11
Update Set-PnPAppSideLoading.md

PowershellScripts:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [ ] Sample
- [x] Parameter description

## Related Issues? ##
N/A

Parameter description


KoenZomers:
I noticed this cmdlet actually no longer exists in PnP.PowerShell. It's from the former module. Wonder if we should add the cmdlet back or remove the documentation.

KoenZomers:
Cmdlet has been taken back, so the documentation can be updated. Thanks @PowershellScripts !

PR #2139, opened 2022-07-13
Removing requirement for https:// prefix with Connect-PnPOnline -Url

KoenZomers:
## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
N/A

Changed that https:// prefix is no longer required when using `Connect-PnPOnline -Url tenant.sharepoint.com` to save some typing. It still supports adding it for backwards compatibility, i.e.  `Connect-PnPOnline -Url https://tenant.sharepoint.com` will still work as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Added
 
 ### Changed
-- Changed to no longer require https:// to be prefixed when using `Connect-PnPOnline -Url tenant.sharepoint.com`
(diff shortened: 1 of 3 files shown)


gautamdsheth:
Hi @KoenZomers , realized that we can't do this because we support page transformation. Now page transformation can happen from OnPrem to Online or Online to Online. In case of Online-to-Online transformation, this would be fine. In case of the OnPrem, there could be cases where the URLs might begin with `http` , so would break those scenarios.

Thoughts ?

KoenZomers:
@gautamdsheth the code change allows http:// and https:// to be used as a prefix, in which case it will not do anything different than it does today. See:

https://github.com/pnp/powershell/pull/2139/files#diff-b885630f1a0a48bcc067ae40d2b76ed6517cdfe18b22450581be1dedbba66ad0R236

PR #2156, opened 2022-07-15
Fix NullDereferenceException in New-PnPSite

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
No linked issues

New-PnPSite uses dynamic parameters, meaning the full set of parameters is determined on the value of -Type . -Type parameter is mandatory, so one would expect it has always a valid value. As explained at https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.idynamicparameters.getdynamicparameters , this is not the case:
>Therefore, parameters which allow input from the pipeline may not be set at the time this method is called, even if the parameters are mandatory.

This means that when invoking `New-PnPSite` with no parameters, the custom dynamic parameters handler will be called before -Type is filled in by the user, causing unexpected behavior in the cmdlet handler:
image

This PR adds some validation to ensure the dynamic parameters handler was successful and an error message that attempts to explain how to call the cmdlet in a working way, with a clearer error message than a null exception:
image

I opted for a validator inside the cmdlet handler as that felt like the cleanest solution, supporting scenarios where the parameter itself is not mandatory. If this PR is approved, I will apply the same to other affected cmdlets:
- Add-PnPTeamsTab
- Set-PnPPage when HeaderType is Custom
- Add-PnPField when Type is Choice
image

All in all I learnt dynamic parameters don't work as one would expect and one should try to avoid them 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Contributors
 
+- Giacomo Pozzoni [jackpoz]
 - James May [fowl2]
 - Jimmy Hang [JimmyHang]
 - Marcus Blennegård [mblennegard]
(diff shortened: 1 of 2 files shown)


KoenZomers:
Thanks @jackpoz !

PR #2162, opened 2022-07-18
Update Restore-PnPFileVersion.md

PowershellScripts:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [X] Sample
- [x] Parameter description

## Related Issues? ##
N/A

Example. Parameter description. Cmdlet description. Moved -Connection parameter to the bottom.


KoenZomers:
There doesn't seem to be any valid scenario where Identity can be left out. Well spotted. I've updated the code and added verbose logging and better error handling while being at it.

PR #2258, opened 2022-08-21
Add check if group is M365 group to the Get-PnPMicrosoft365Group command

reusto:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

Added check if the group is a Microsoft 365 group when getting a group by ID with the Get-PnPMicrosoft365Group command.
This is to align the return of the command with `Get-PnPMicrosoft365Group | Where-Object {$_.Id -eq '{guid}'`

~~When a group with the specified ID exists that is not a Microsoft 365 group, then an error will be thrown including a hint to use the `Get-PnPAzureADGroup` command.~~

**Update 24.08.2022:**

Hint regarding usage of `Get-PnPAzureADGroup` command dropped due to different implementation.

Added fixes for the following bugs:

- Get-PnPMicrosoft365Group returns a group if the mailNickname of the group matches the provided Identity even if the group is not a Microsoft 365 group
- Get-PnPMicrosoft365Group doesn't return a group if the format of the provided identity matches the GUID format but is meant to be the display name or mail nickname

### DIFF
--- a/src/Commands/Utilities/Microsoft365GroupsUtility.cs
+++ b/src/Commands/Utilities/Microsoft365GroupsUtility.cs
@@ -7,6 +7,7 @@ using PnP.PowerShell.Commands.Utilities.REST;
 using System.Text.Json;
 using PnP.PowerShell.Commands.Model;
 using PnP.PowerShell.Commands.Base;
+using System.Management.Automation;
 
 namespace PnP.PowerShell.Commands.Utilities
 {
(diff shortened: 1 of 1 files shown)


gautamdsheth:
hi @reusto , can you do something like this at API level itself instead of the results ? 

`.../v1.0/groups?$filter=id eq '02eea2f2-9cf9-4cac-80fb-07be37177f72' and groupTypes/any(c:c+eq+'Unified')`

We can then check if its null or not. 
If null, we can add a warning/error then ? 
Would love to hear your thoughts on this 😊

reusto:
hi @gautamdsheth , technicall it is possible to do this at the API level as you suggest.

I like your suggestion due to multiple reasons:

- Performance seems to be similar to my original implementation.
- It allows for code consolidation with the GetGroupAsync function when using display name.
- It allows to have edge cases in mind like a user creating a group with a display name or mailnickname in the format of a GUID and then trying to get the group by display name or mail nickname, which is why I extended the filter query to include display name and mail nickname even when providing an input that seems like a GUID to 
`.../v1.0/groups?$filter=groupTypes/any(c:c+eq+'Unified') and (id eq '{groupId}' or displayName eq '{groupId}' or mailNickName eq '{groupId}')`

I already implemented this and created a new commit.

The commit also includes fixes for the following bugs that I found while looking into your suggestion:

- Get-PnPMicrosoft365Group returns a group if the mailNickname of the group matches the provided Identity even if the group is not a Microsoft 365 group (this was due to missing brackets in the filter query when getting the group by displayname or mailnickname)
- Get-PnPMicrosoft365Group doesn't return a group if the format of the provided identity matches the GUID format but is meant to be the display name or mail nickname

I used the implementation of the previous GetGroupAsync (by displayName) function with additions from the GetGroupAsync (by groupId) function, but also changed the behavior when no group is returned to throw an error. This is to match with the previous behavior of the GetGroupAsync (by groupId) function and to be consistent with the behavior of other commands in this module which also throw an error when an item with the provided identity can't be found. (for example Get-PnPListItem or Get-PnPList).

Due to the new implementation the hint regarding the usage of Get-PnPAzureAdGroup was dropped as we now can't differentiate wether a group is not returned because it doesn't exist or because it is not a Microsoft 365 group. But this shouldn't be a problem as it isn't relevant for the command and was previously just implemented as a nice to have.

gautamdsheth:
Thank you so much @reusto ! 
Much appreciated 😊🙏

PR #2303, opened 2022-09-01
AzureADAppSitePermission: improve completion of "Permissions" parameter

fowl2:
Adds missing value "owner". 

"Manage" and "FullControl" aren't listed in the docs, but I kept them just in case.

Switching to `EnumAsStringArgumentCompleter` to allow any newly added value to be used without a code change but to still suggest the known ones.


KoenZomers:
@fowl2 Its not clear to me what switching to ArgumentCompleter adds, even after reading a blog on it. Can you elaborate what you feel that it adds?

The API behind this is a bit odd. When creating a new app permission, you can only choose Read and Write. Nothing else. When updating the same permission, you can choose from Read, Write, Manage and FullControl. I've checked this with the devs of this API and there is no way to initially use these other two values. The Owner option as specified in the API I haven't heard of before and when testing with it, it doesn't work.

I will update the PR to use the enum but create two versions of it. One for adding and one for updating, so it reflects what is technically possible.

Please elaborate on the ArgumentCompleter as it will make us deviate from how other cmdlets are set up. If it really adds true value, we may want to do it for others as well.

fowl2:
The benefit of the completer over a validator is that it allows new possible values to be used without updating the cmdlet. 

eg. if the service grows support for 'ContributeWithoutDelete', the cmdlet could still be used, just without the completion. Basically "we suggest one of these values, but will let you try anything else". It makes the most sense for potentially open ended extensibility points with some common values or hotly anticipated/fast moving parameters.

On the API itself, that does seem very odd and if that's the case your changes LGTM :) I made a comment in https://github.com/microsoftgraph/microsoft-graph-docs/issues/14178 about updating the docs.

fowl2:
If new values are not expected - a slight alteration of this PR would be to use the built-in PowerShell completion and validation logic by changing the types of the parameters from `string` to the `enum` types created here.

PR #2308, opened 2022-09-01
Feature: added support for shared channel creation

gautamdsheth:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
NA

Added support for shared channel creation.
Made -Private parameter obsolete, instead will use TeamMemberShipType parameter.

For some reason, the API is throwing unknown error at the moment, am I missing something ?


Studermarc:
@gautamdsheth Wouldn't it be better to use "Standard" instead of "Public" for the standard channels? This way it is also aligned with how Microsoft documents the different channel types. "Public" might lead to some confusions, as it doesn't appear in any Microsoft documentation.

KoenZomers:
Fair point @Studermarc, I have updated the code according to your suggestion

@gautamdsheth fixed some bugs with the paramsets, tested the cmdlet, updated the documentation and added a changelog entry. If you agree with these changes, go ahead and merge.

gautamdsheth:
Looks good to me !

PR #2342, opened 2022-09-13
Fix build warning from duplicate PackageReference for 'NewtonSoft.Json' 

fowl2:
(The version stays the same.)


gautamdsheth:
Thanks @fowl2 !

PR #2354, opened 2022-09-18
Feature #2249 - add support for managed identity for SPO

### LINKED ISSUE #2249: [FEATURE] Add ability to use Managed Identity with SharePoint Online
The documentation of PnP.PowerShell describes that ManagedIdentity cannot be used when executing commands against SharePoint:

https://pnp.github.io/powershell/cmdlets/Connect-PnPOnline.html?q=ManagedIdentity#-managedidentity

However: I wonder why this is, as I can access SharePoint just fine in my C# Functions, using a combination of Azure Identity and PnP Framework.

It would be awesome if we could add this, so that people can execute their scripts without having to register an Azure AD app with certificate authentication.

gautamdsheth:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #2249 

Add support to use Managed identity for SPO APIs besides Graph


KoenZomers:
@gautamdsheth I've checked the code and have done bugfixes to it. What I cannot get to work yet is that if you have multiple permissions set in your system identity token:

image

It will still only get an access token containing the scp for the first one you have configured for that source (i.e. Graph or SPO). Not for the other ones. I also cannot figure out how to request a specific scope. The .default trick also does not seem to work.

PR #2381, opened 2022-09-26
SetList: add "-Path" parameter for updating list URLs

fowl2:
Add `-Path` parameter for updating list URLs.

Considered calling it `-Url` but that would seem more appropriate if full URLs were expected to be the common form.

Looked into `Folder.MoveToPath`, but I think the built-in URL encoding/decoding guessing makes more sense here. "Shared Documents" and "Shared%20Documents" should both be "Shared%20Documents" rather than the later becoming "Shared%2520Documents". 

```powershell
Set-PnPList -Identity "Demo List" -Path "Lists/DemoList"
Set-PnPList -Identity "Shared Documents" -Path "Documents"
```

## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample


KoenZomers:
Thanks @fowl2 !

Issue #2018, opened 2022-06-21
[FEATURE] Set-PnPListItem without values/always Update()


fowl2:
**Is your feature request related to a problem? Please describe.**
Often when developing or troubleshooting issues I need to trigger an (remote) event receiver, workflow, flow or webhook without changing any field values. 

Currently `Set-PnPListItem` will not call the appropriate *`Update()` method if no field values or content type are specified or null/empty. I'm not sure if this is intentional, but it's a nice slight optimization if eg. the `Values` `Hashtable` is empty - the number of .

**Describe the solution you'd like**
Addition of a new parameter like `-AlwaysUpdate`, `-Force`, `-AllowEmpty`, etc.

**Describe alternatives you've considered**
1. Always *`Update()` if no other parameters are specified. Might be a bit subtle that `Set-PnPListItem` and `Set-PnPListItem -Values @{}` are different.
2. Always *`Update()` if `UpdateType` is specified.
3. Any of the above, but add a parameter to do in the inverse - ie. `-SkipUnchanged` (name to be bikesheeded).
4. A new cmdlet like `Touch-PnPListItem`. Discoverability might be an issue as `Touch` isn't on the [approved verbs list](https://docs.microsoft.com/en-us/powershell/scripting/developer/cmdlet/approved-verbs-for-windows-powershell-commands) 
5. Some combination of above.

**Additional context**
I'd be happy to implement this feature.

--------

Other notes:
1. It's not clear what `Set-PnPListItem -List $list -Identity $id -ContentType:$null` means, but it doesn't currently cause an error - just basically the same thing as `Get-PnPListItem -List $list -Id $id`.
2. According to the approved verbs list apparently `Update` is apparently not recommended for this cmdlet either - it should be `Edit`. Not sure I agree with that, but maybe I'm just used to it now.

gautamdsheth:
@fowl2 - I guess we can add a `-Force` additional parameter for this. That would make it consistent with other cmdlets.

gautamdsheth:
closing it as PR has been merged.

PR #2396, opened 2022-09-29
Set-PnPListItem: Add -Force (and other fixes)

fowl2:
Adds a `-Force` parameter to `Set-PnPListItem` to allow forcing of updates even when not updating the item (https://github.com/pnp/powershell/issues/2018). Also reducing and whitespace fixes.

Also includes separate commits for these other issues I noticed:
* Fix -SystemUpdate:$false
* Throw when item can't be found and warn when content type can't be found

Can split out if required but risk merge conflicts with above.

Hope this helps :)


documentation/Set-PnPListItem.md
```diff
@@ -283,6 +283,21 @@ Accept pipeline input: False
 Accept wildcard characters: False
 ```
 
+### -Force
+
+Forces the update of the list item even if there are no value changes.
```
KoenZomers, line 288: Could you elaborate in this help text on why this would be useful, like you already explained in the issue? Will help others wondering why they would ever need this.

fowl2: how's that?

KoenZomers: Perfect!

KoenZomers:
Thanks @fowl2 !

fowl2:
@KoenZomers nooo `-SystemUpdate` is my friend! 😥😥😥😥

KoenZomers:
> @KoenZomers nooo `-SystemUpdate` is my friend! 😥😥😥😥

No worries, `-UpdateType SystemUpdate` remains at your disposal to do exactly that

fowl2:
I guess it's a personal thing, but I use `-SystemUpdate` all the time so the extra typing is something I'm not particularly enjoying.

SjoerdV:
> I guess it's a personal thing, but I use `-SystemUpdate` all the time so the extra typing is something I'm not particularly enjoying.

I agree, and `Set-PnPFolderPermission` still uses `-SystemUpdate`, so there might be some inconsistency added by this change. I vote to return to just using `-SystemUpdate`

BTW I am using 2.1.1 and `-SystemUpdate` still is a parameter for `Set-PnPListItem`, seems like the change was undone, without specifying this in the release notes for any release after 1.12.0 perhaps? Weird ;-)

PR #2409, opened 2022-10-02
Fix Set-PnPList -MinorVersion documentation

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
No related issues

`Set-PnPList -MinorVersion` affects CSOM property [MajorWithMinorVersionsLimit](https://learn.microsoft.com/en-us/previous-versions/office/sharepoint-server/mt162529(v=office.15)), described by SharePoint with "Keep drafts for the following number of major versions:" 
https://github.com/pnp/powershell/blob/7f498aa5ddd2d18784de026b11f4824681e973d9/src/Commands/Lists/SetList.cs#L251
image

The previous documentation stated "Maximum minor versions to keep" but SharePoint doesn't support to specify how many minor versions to keep, that's a fixed limit of 511. See https://support.microsoft.com/en-us/office/how-versioning-works-in-lists-and-libraries-0f6cd105-974f-44a4-aadb-43ac5bdfd247 for official documentation:
>The maximum number of minor versions is 511, and the number of major versions can be adjusted.

It might be good to rename the parameter `-MinorVersion` to something else, maybe using the same name as CSOM `-MajorWithMinorVersionsLimit`, but that will break all existing scripts. For now fixing the documentation should be good enough.

Feel free to change the wording, I'm not a native english speaker :)


gautamdsheth:
Thanks @jackpoz !

jackpoz:
> It might be good to rename the parameter `-MinorVersion` to something else, maybe using the same name as CSOM `-MajorWithMinorVersionsLimit`, but that will break all existing scripts. For now fixing the documentation should be good enough.

@gautamdsheth any opinion about this ? Would it be good to add `-MajorWithMinorVersionsLimit` as parameter alias and add a deprecation notice to `-MinorVersion` , mentioning that it might have a different behavior than what users have always expected ? Then maybe after enough time removing the `-MinorVersion` alias

PR #2425, opened 2022-10-05
Set-PnPTeamsChannel: IsFavoriteByDefault 'Nullable object error  (#2424)

### LINKED ISSUE #2424: [BUG] Set-PnPTeamsChannel: IsFavoriteByDefault 'Nullable object must have a value.'
If “Automatically show this channel in everyone’s channel list” (or in code “IsFavoriteByDefault”) is set to true, it is also saved as true but when it is set to false, NULL is saved.
 
When this “IsFavoriteByDefault” is NULL and the Set-PnPTeamsChannel function is triggered the execution of the commandlet will fail with a 'Nullable object must have a value.' Error.

### Expected behavior 
Updating the teams channel by calling the Set-PnPTeamsChannel function.

### Actual behavior 
'Nullable object must have a value.' Error.

Set the “IsFavoriteByDefault” to false (UI, MS Graph or PS) (NULL will be saved). 
After this, the property is set to NULL.
When the Set-PnPTeamsChannel function is called again it will fail with a 'Nullable object must have a value.' Error. 
This happens because “teamChannel.IsFavoriteByDefault” is NULL and the following code is called: “teamChannel.IsFavoriteByDefault.Value”

1.11.0

- [x] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

RobLempens:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #2424 

Removing the .Value of the nullable object  "IsFavoriteByDefault" to avoid the "Nullable object must have a value" error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,8 +62,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed `Import-PnPTermGroupFromXml` issue where a valid template was not working. [#2353](https://github.com/pnp/powershell/pull/2353)
 - Fixed `Set-PnPTenant` cmdlet not working when `-Force` parameter is specified. [#2373](https://github.com/pnp/powershell/pull/2373)
 - Fixed `Add-PnPTeamsTab` cmdlet not working with certain types when using dynamic parameters. [#2405](https://github.com/pnp/powershell/pull/2405)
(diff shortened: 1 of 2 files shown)


KoenZomers:
Thanks @RobLempens !

PR #2452, opened 2022-10-11
Add support to Remove-PnPSiteCollectionAppCatalog to remove app catalog by Site Id

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
No related issues

Add support to Remove-PnPSiteCollectionAppCatalog to remove app catalog by Site Id by calling `SiteCollectionAppCatalogsSites.RemoveById(guid)` instead of `Remove(url)`. This is needed when a site with an app catalog has the Url renamed or gets deleted.
Prefer `RemoveById(guid)` method when passing

New behavior with deleted site, the app catalog can be deleted:
image

Old behavior with deleted site, an error is thrown:
image

Old behavior with site deleted and recreated, the app catalog is not deleted:


gautamdsheth:
Thanks @jackpoz, merged it ! 

Awesome find, thanks for sharing this 😊🙏

PR #2472, opened 2022-10-16
Feature/recyclebin paging

koskila:
Implement paging for getting recycle bin items.

## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #1598

## What is in this Pull Request ? ##
Abstracts getting recycle bin items to a new helper class ("utility) and implements paging. The code is a bit verbose, but I hope you can deal with it - I can of course work on it further if need be :)

Note - I had to remove the ordering of the items (by deleted date, desc) to make paging work. It could be implemented on the client-side, if we want to?

PR #2473, opened 2022-10-16
Docs improvements

koskila:
Some small documentation improvements that I came upon while working on other stuff. Fixes a few typos and clarifies some commands where I thought it could be useful :)

## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
If #2472 is ok, I could also add a sentence about the paging (it slows the queries down a bit, since executing multiple synchronous requests is slower than executing just one large - but I switched to this model as it's more likely not to go over the list view threshold).

Updates to documentation.


gautamdsheth:
Thank you @koskila, much appreciated 🙏😊

koskila:
My pleasure!

Issue #2485, opened 2022-10-20
[BUG] Restore-PnPRecycleBinItem -RowLimit parameter appears to be broken


NoahL989:
Hello,

It appears the issue #1598 is not fixed and this is a request to reopen it.

Running into the same issue as #1598 where the -RowLimit parameter appears to not work for Restore-PnPRecycleBinItem.
![image](https://user-images.githubusercontent.com/116316586/197073357-89fec206-9161-4b91-b02c-799f84b1c548.png)

Even when running the command without the filtering it errors.
![image](https://user-images.githubusercontent.com/116316586/197073496-35d5d91e-415f-4f3d-ae5f-033b6e712eb3.png)

### Reporting an Issue or Missing Feature
It is an Issue.

### Expected behavior 
For the cmdlet to work as intended.

### Actual behavior 
Getting "The attempted operation is prohibited because it exceeds the list view threshold" error when trying to restore something and the -RowLimit parameter does nothing to fix the error.

### Steps to reproduce behavior
Connect-PnPOnline -Url https://example.sharepoint.com/sites/sharepointsite -UseWebLogin
Get-PnPRecycleBinItem -RowLimit 15000 | ? -Property DeletedByName -Like "Firstname Lastname" | Restore-PnPRecycleBinItem -Force -RowLimit 15000

OR

Connect-PnPOnline -Url https://example.sharepoint.com/sites/sharepointsite -UseWebLogin
Restore-PnpRecycleBinItem -Identity 22c4d077-2a3d-446e-bf30-e9ece79820a1 -Force -RowLimit 1

### What is the version of the Cmdlet module you are running?
(you can retrieve this by executing ```Get-Module -Name "PnP.PowerShell" -ListAvailable```)

Running version 1.11.109
![image](https://user-images.githubusercontent.com/116316586/197073816-a473947f-1561-4638-81db-6c52be3db08d.png)

### Which operating system/environment are you running PnP PowerShell on?
- [X] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

koskila:
This is fair feedback - I implemented the fix to Get-PnPRecycleBinItem, where it should now work, and tested by piping (which should work but will be incredibly slow).

The current implementation of Restore-PnPRecycleBinItem only works with an Identity of some sort, it ignores RowLimit and will never reach the part of code where it would try restoring everything.

Let me take a look at how to improve this...

[...]

PR #2499, opened 2022-10-27
Super basic simple fix: now accepts -RowLimit parameter

koskila:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Partial fix to #2485.

The commandlet now accepts RowLimit parameter (earlier it was always ignored due to how Identity was managed, and the commandlet could effectively be used to only restore 1 item at a time). This implementation blindly accepts the -RowLimit (without validating whether the number makes sense).

A proper fix should also implement paging or at least verify that restoring with high RowLimits works (which I can also test myself later on if need be).

Also adds a nicer message for confirming restoring X number of items.

### DIFF
--- a/src/Commands/Properties/Resources.Designer.cs
+++ b/src/Commands/Properties/Resources.Designer.cs
@@ -725,6 +725,15 @@ namespace PnP.PowerShell.Commands.Properties {
         }
         
         /// <summary>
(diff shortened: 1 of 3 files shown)


gautamdsheth:
Thank you @koskila, much needed fix.

PR #2503, opened 2022-10-27
Feature: support uploading image for Image column

gautamdsheth:
## Type ##
- [x] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
NA

Supports uploading image to the site assets library for image column


KoenZomers:
@gautamdsheth What will we do with this one for the next release as I believe we discussed this implementation is quite minimalistic?

gautamdsheth:
@KoenZomers - I guess we can squeeze it in for this release ?

KoenZomers:
@KoenZomers Don't you think we'll make it difficult for ourselves to evolve on this in a future version with backwards compatibility if we do? If you think we can work our way through that, let's merge and see what kind of feedback we receive on shortcomings.

gautamdsheth:
@KoenZomers - we should be ok now. Was a bit difficult in current format because we were reliant on existing uploaded files. The parameter `-ServerRelativePath` makes more sense IMHO for that scenario. We can use `-Path` parameter for file to be uploaded from the local file system, which makes more sense.

Thoughts?

KoenZomers:
Let's do it then

PR #2607, opened 2022-12-04
Fix setting multi taxonomy field value when input is PSObject[]

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
None

Fix setting multi taxonomy field value when input is PSObject[] .

Using the script below, an error is generated when copying TaxonomyFieldTypeMulti field values from a list item to another, storing them in a temporary array:
```powershell
# Copy multi taxonomy field values from list item ID 1 to list item ID 2
$list = get-pnplist "List with taxonomy"
$item = get-pnplistitem $list -Id 1

# Add each term to a temporary array
$values = @()
$item.FieldValues.Multi.TermGuid | %{$values += $_}

# Update list item 2
Set-PnPListItem -Identity 2 -Values @{Multi = $values} -List $list
$values
$values.GetType().FullName
$values[0].GetType().FullName
```
image

This is caused by the C# code expecting the input to be of type String but PowerShell wraps it around a PSObject when using Foreach-Object at `$item.FieldValues.Multi.TermGuid | %{$values += $_}`
image

Instead of doing a cast, this PR changes the code to just do a ToString().
See below the same code, this time not throwing any exception:
image

image

### DIFF
--- a/src/Commands/Utilities/ListItemHelper.cs
+++ b/src/Commands/Utilities/ListItemHelper.cs
@@ -114,10 +114,10 @@ namespace PnP.PowerShell.Commands.Utilities
                                     {
                                         TaxonomyItem taxonomyItem;
                                         Guid termGuid;
-                                        if (!Guid.TryParse(arrayItem as string, out termGuid))
+                                        if (!Guid.TryParse(arrayItem?.ToString(), out termGuid))
                                         {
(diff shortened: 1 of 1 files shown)


KoenZomers:
Thanks for the contribution @jackpoz !

PR #2646, opened 2022-12-13
Added `Get-PnPFolderStorageMetric`

robinmeure:
Added Get-PnPFolderStorageMetric

## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

Added functionality to retrieve storage metrics per folder using the PnPCore adjusted way of retrieving these.
This is an attempt to get the same information as the storman.aspx page gives, but only per folder (e.g. document library)


KoenZomers:
Only documentation needs to describe the difference between these two output variables:

image

Other than that, looks good for merge.

robinmeure:
It's bytes, one is the size of the document itself (TotalFileStreamSize) and if the document has more than one version, the total size it consumed is reflected in TotalSize.

khha:
@robinmeure Thanks, I found also the command "Measure-PnPList". I see differance output. Do you know what is the differance between "Get-PnPFolderStorageMetric" and "Measure-PnPList" ?

image

image

PR #2687, opened 2022-12-30
Feature: Added support for HTTP/2 to improve perf

gautamdsheth:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
NA

Added `HTTP/2` support for all HTTP requests to improve performance.
It will by default fallback to `1.1` if the server doesn't support.


KoenZomers:
Tested this PR. When I execute i.e. `Get-PnPMicrosoft365Group`, I see it has HTTP 2.0 in the requestmessage. Fiddler however shows a normal HTTP 1.1 connection is being made:

image

image

Is that as expected? What would make the code ask the server for HTTP 2.0 I'm wondering?

KoenZomers:
No issues from my current machine. Must have been a connection issue. Going to proceed with merging.

PR #2727, opened 2023-01-18
Fixed Debug build dependencies

stokuri:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

1. Fix for source directory location (no more required to be under the drive root).
2. Fix references when PnPCoreSdkPath is specified. Always add local libraries as reference.


gautamdsheth:
Hello @stokuri , what exactly it is that you are trying to fix here ? 
The current setup works perfectly fine for the standard dev tasks, but maybe we are missing some scenario ?

KoenZomers:
Thanks for filing this separately @stokuri. We're still not clear on what you're trying to fix with this :) Can you elaborate what scenario you're in using what doing what where things fail today which this would fix so we can better understand why a change would be needed?

KoenZomers:
Havent heard back from you @stokuri. As its not clear to us what the intention is, we've decided to close it for now. If you wish to resume the conversation, feel free to respond.

PR #2736, opened 2023-01-20
Changes to `Add-PnPNavigationNode`

### LINKED ISSUE #2708: [BUG] AudienceIds not working after running Add-PnPNavigationNode. This has been an issue for 4 years but no one us taking responsibility.
Issue

### Expected behavior 
Please let me know if this isn't the right place for this question/bug report. When I run the following code (see below) to add a Quick Launch link and add an audience it seems to work just fine. No errors are thrown and when I go to the SharePoint site I can see the audience associated with the guid was added to the link.

The problem is that the Quick Launch link is still visible to me even though I am not part of the audience. I was expecting the link to be hidden. Even more strange is that if I add the audience to the link using the UI on the SharePoint site it works as expected. The link is hidden from me.

Is there something I'm not doing correctly? All the websites I can find online say this code should work.

Almost 4 years ago someone had the same issue: https://sharepoint.stackexchange.com/questions/260831/how-to-set-navigationnode-audienceids

### Actual behavior 
Here is a screen shot. The Audience Test 9 link should not be visible to me because I am not in that group. Even after refreshing the page it is still visible.
image

```
$newNode = Add-PnPNavigationNode -Location "QuickLaunch" -Title "Audience Test" -Url "http://linkless.header/"

$audienceList = New-Object System.Collections.Generic.List[guid]
$audienceList.add([System.guid]::New(("7e5c22f2-1254-453e-84f4-58ce4ed3e995")))

$newNode.AudienceIds = $audienceList
$newNode.Update()
Invoke-PnPQuery
```

1.11.0

- [X ] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

KoenZomers:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #2708 

- Fixed issue with `Add-PnPNavigationNode` not always showing the new navigation node without requiring a manual edit and save of the menu first
- Added `-AudienceIds` option to `Add-PnPNavitationNode` to allow setting the audience for a navigation node


gautamdsheth:
@KoenZomers - I would have approved this PR yesterday but then I attended the PnP community call where Bert mentioned that he added Audience targeting support in PnP Core SDK 😊

Can we maybe use that ? 

Check this link: 

https://pnp.github.io/pnpcore/using-the-sdk/audience-targeting-intro.html#in-navigation

If you want, I can write a separate PR for that.

KoenZomers:
Interesting. He [seems to create the JSON himself](https://github.com/pnp/pnpcore/blob/974d93907f700028f99250c50b29a7b8396acd9f/src/sdk/PnP.Core/Model/SharePoint/Settings/Internal/NavigationNode.cs#L108) and post that to the new endpoints. Would make things more efficient, but needs testing if it works well. Wish I would have known this before. Would have saves me a few hours of getting it to work this way :)

PR #2786, opened 2023-02-06
Update AddTeamsChannel.cs -- OwnerUPN shouldn't be mandatory in the Specific Parameter Set

Studermarc:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Standard Channels cannot be created with the "ChannelType" Parameter

Changed the "OwnerUPN" to Mandatory = false for the "Specific" Parameter Set. Setting this Parameted to Mandadory is preventing Standard Channels to be created with the "ChannelType" Parameter, as it will always ask for an OwnerUPN, whereas none should be required to begin with.

Also correcting a small typo in the obsolete message for the "Private" switch.


KoenZomers:
Thanks for your contribution @Studermarc . The typical scenario was that when wanting to create a public/standard channel, one would just leave `-ChannelType` out, which would make it default to this type and not require an OwnerUPN to be specified. But indeed, this change makes sense so that if someone would provide it, it will still work. Merging your proposal with some small adjustments.

Studermarc:
Awsome, many thanks.

PR #2811, opened 2023-02-13
Feature: reduce dependency of Graph SDK in our cmdlets

gautamdsheth:
## Type ##
- [x] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
Fixes NA

Removed code that depends on Graph SDK for select cmdlets.

Had to add a nullable enum for Group Visibility property because it can be nullable for some Azure AD groups and it caused issues in JSON serialization/de-serialization.


KoenZomers:
As dicussed. These class files still have dependencies towards the MS Graph SDK and still would need fixing before it can entirely be removed:

image

PR #2821, opened 2023-02-14
Add connection paramter to Connect-PnPOnline

reusto:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

Added a connection parameter to the Connect-PnPOnline command to allow the reuse of a specific connection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added support for specifying the ContentUrl and WebsiteUrl configuration in `Add-PnPTeamsTab` cmdlet when trying to add a SharePoint page or list as a tab in Teams channel. [#2807](https://github.com/pnp/powershell/pull/2807)
 - Added `-CheckinType` parameter to `Add-PnPFile` cmdlet which provides the option to specify the checkin type for a file. The default value is set to `MinorCheckIn`. [#2806](https://github.com/pnp/powershell/pull/2806)
 - Added `-ApplicationId` as alias for `-ClientId` in `Connect-PnPOnline` and `Request-PnPAccessToken` cmdlets. [#2805](https://github.com/pnp/powershell/pull/2805)
+- Added `-Connection` option to `Connect-PnPOnline` which allows of reusing an authenticated connection to connect to a different site [#2821](https://github.com/pnp/powershell/pull/2821)
 
 ### Changed
 
@@ -78,6 +79,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Contributors
 
+- [reusto]
 - Ronald Mavarez [ronaldmavarez]
 - [lilealdai]
 - Martin Lingstuyl [martinlingstuyl]
--- a/documentation/Connect-PnPOnline.md
+++ b/documentation/Connect-PnPOnline.md
@@ -19,14 +19,14 @@ Connect to a SharePoint site
 Connect-PnPOnline [-ReturnConnection] [-Url] <String> [-Credentials <CredentialPipeBind>] [-CurrentCredentials]
  [-CreateDrive] [-DriveName <String>] [-ClientId <String>] [-RedirectUri <String>]
  [-AzureEnvironment <AzureEnvironment>] [-TenantAdminUrl <String>]
- [-TransformationOnPrem] [-ValidateConnection]
+ [-TransformationOnPrem] [-ValidateConnection] [-Connection <PnPConnection>]
 ```
 
 ### SharePoint ACS (Legacy) App Only
 ```
 Connect-PnPOnline [-ReturnConnection] [-Url] <String> [-Realm <String>] -ClientSecret <String> [-CreateDrive]
  [-DriveName <String>] -ClientId <String> [-AzureEnvironment <AzureEnvironment>] [-TenantAdminUrl <String>]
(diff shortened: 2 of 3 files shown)


KoenZomers:
Interesting one. I guess the use case for this would be if you wanted to connect to a different site URL but not having to authenticate again, correct?

reusto:
Correct, that is already the case if the current connection is stored in PnPConnection.Current, but if I save the connection to a variable by using -ReturnConnection this value is not set and can't be reused.

For me for example this is useful when using parallel threads to iterate through multiple site collections while having to use a connection method like interactive.

KoenZomers:
Thanks @reusto !

MubSalim:
Hi, 

The current Connect-PnPOnline documentation lists a -Connection parameter: 
https://pnp.github.io/powershell/cmdlets/Connect-PnPOnline.html#-connection

but in PnP.PowerShell v1.12.0 I can't see that parameter. Is this a case of the documentation having been updated before the command is fully available or am I missing something. 

image

Thanks in advance for the clarification!

LastGunslinger:
@KoenZomers  I am trying to use this new feature in PnP.PowerShell 2.1.1, and it doesn't seem to be working for me. Perhaps I'm doing something wrong or don't understand how it works? This was my simple test:

```powershell
$creds = Get-Credential
$conn1 = Connect-PnPOnline 'tenant.sharepoint.com/sites/site1' -Credentials $creds -ReturnConnection
$conn2 = Connect-PnPOnline 'tenant.sharepoint.com/sites/site2' -Connection $conn1
```

My issue is, when I run the 3rd line it asks for my credentials again. It was my understanding that the "-Connection" parameter would allow creating the new connection without also needing to provide the credentials a second time, otherwise I don't see what problem this is meant to solve.

KoenZomers:
@LastGunslinger, that's a bit of a quirky scenario. The change causes it now to be possible to use different authentication methods and parameters to allow for scenarions where you would want to connect using different credentials\app regs\certs\etc. However it does cause the reuse of credentials not to work this straightforward anymore as each connection is considered as a connection on its own. To mitigate this to have it do what you would like it to do, what you can do is first connect using the "normal" way, i.e.:

```PowerShell
Connect-PnPOnline tenant.sharepoint.com/sites/site1 -Credentials $creds
```

Once you've done this, you should be able to create additional connections using i.e.

```PowerShell
$conn1 = Connect-PnPOnline tenant.sharepoint.com/sites/site2 -ReturnConnection
```

And it should reuse the credentials from the initial connection. Give it a try and see if it works this way for you.

LastGunslinger:
> @LastGunslinger, that's a bit of a quirky scenario. The change causes it now to be possible to use different authentication methods and parameters to allow for scenarions where you would want to connect using different credentials\app regs\certs\etc. However it does cause the reuse of credentials not to work this straightforward anymore as each connection is considered as a connection on its own. To mitigate this to have it do what you would like it to do, what you can do is first connect using the "normal" way, i.e.:
> 
> ```powershell
> Connect-PnPOnline tenant.sharepoint.com/sites/site1 -Credentials $creds
> ```
> 
> Once you've done this, you should be able to create additional connections using i.e.
> 
> ```powershell
> $conn1 = Connect-PnPOnline tenant.sharepoint.com/sites/site2 -ReturnConnection
> ```
> 
> And it should reuse the credentials from the initial connection. Give it a try and see if it works this way for you.



@KoenZomers I tried to connect using the same method that you did above, and I am still being prompted to enter my credentials again for the second connection. This was my test:

```powershell
Connect-PnPOnline tenant.sharepoint.com/sites/site1 -Credentials $adminCreds
$conn = Connect-PnPOnline tenant.sharepoint.com/sites/site2 -ReturnConnection
```
After the second line, it again prompted me for a username/password for the connection. Just to be sure that I was indeed using the correct PnP.PowerShell version, I ran `Get-Module -Name PnP.PowerShell` to make sure that the currently loaded version was 2.1.1, and it was, so I'm not sure why this isn't working for me.

reusto:
@LastGunslinger I can see the reason for this behavior in the code. Therefor yes currently you need to provide the credentials additionally to the connection parameter. 

@KoenZomers If its okay with you I will take a look into how to better navigate the reuse of a connection when just providing the connection.

KoenZomers:
You always need to provide a means of authentication indeed. Not providing anything will always make it prompt for credentials.

Feel free indeed to see how this can be improved. Was thinking about adding a flag like -reuseauthentication or something similar. Just mind to keep the option open to actually be able to use a different authentication.

KoenZomers:
As the matter of fact, what is particularly interesting about this specific scenario is, you seem to be authenticating with just an username and password so without mfa. I don't see how that could be a problem in any scenario actually as you capture the credentials one using Get-credentials and then just pass them in with every connect. What am I missing here?

LastGunslinger:
> As the matter of fact, what is particularly interesting about this specific scenario is, you seem to be authenticating with just an username and password so without mfa. I don't see how that could be a problem in any scenario actually as you capture the credentials one using Get-credentials and then just pass them in with every connect. What am I missing here?


There's not really an issue here for me, which is why I didn't open this as a bug. When I do have to connect to many sites in separate threads, I just store the credentials object and re-use that, like you mentioned above. I just wasn't sure if what I was experiencing was intentional, since it didn't seem to match up with what I saw in the change log. I'm fine with continuing to re-use the credentials, I was just testing this new functionality to see if it would save any time when running my scripts. With thousands of sites, even a savings of a few seconds can make a difference :)

PR #2827, opened 2023-02-16
installation.md: add " -Scope CurrentUser"

fowl2:
add " -Scope CurrentUser" to increase install success rate by not requiring administrative rights


KoenZomers:
Good idea, thanks @fowl2

PR #2873, opened 2023-03-07
Added `-OpenDocumentsIn` option to `Set-PnPList`

KoenZomers:
## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
N/A

Added `-OpenDocumentsIn` option to `Set-PnPList` which allows configuring if documents should be opened in the browser or in the local client


gautamdsheth:
Can we use `-OpenDocumentsMode` as parameter name ? Maybe also add an additional check to determine if the list is actually document library , I think the template enum was `101`, if not add warning this wont work ?

KoenZomers:
I now see we're actually using the doclib check already, so I guess it does work well:
https://github.com/pnp/powershell/blob/890e4c15af8702ed628c39a7d7687768a5700ea6/src/Commands/Lists/SetList.cs#L244

Let me add it for this PR too then.

PR #2881, opened 2023-03-10
Feature #2863: Added support for setting the Azure AD app logo if specified

### LINKED ISSUE #2863: [FEATURE] Register-PnPAzureADApp to include setting the app logo
No, it's not a problem per se. I just think that the use of a logo makes the app appear more recognizable.

Add a parameter and underlying code to Register-PnPAzureADApp to support setting the logo for the app that will be registered.
The parameter can reference a file path to an image file containing the logo

Manual configuration of the logo is possible, but comes with an additional stap opposed to a complete automation of creating and configuring the registration.
Screenshot 2023-03-03 121418

**Additional context**

gautamdsheth:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #2863 

Added support for setting the Azure AD app logo if specified.


src/Commands/AzureAD/RegisterAzureADApp.cs
```diff
@@ -637,5 +645,52 @@ private void StartConsentFlow(string loginEndPoint, AzureADApp azureApp, string
...
+                {
+                    WriteVerbose("Setting the logo for the Azure AD app");
+
+                    var endpoint = $"https://{AuthenticationManager.GetGraphEndPoint(AzureEnvironment)}/v1.0/applications/{azureApp.Id}/logo";
+
+                    var bytes = File.ReadAllBytes(LogoFilePath);
+                    var byteArrayContent = new ByteArrayContent(bytes);
+
+                    var mimeType = "";
+                    if (LogoFilePath.EndsWith("gif", StringComparison.InvariantCultureIgnoreCase))
```
KoenZomers, line 667: Wouldn't it be an idea to run the path through a FileInfo instance so you can use the .Extension attribute, which would be a safer/cleaner approach, in my opinion?

src/Commands/AzureAD/RegisterAzureADApp.cs
```diff
@@ -637,5 +645,52 @@ private void StartConsentFlow(string loginEndPoint, AzureADApp azureApp, string
...
+                    }
+                    if (LogoFilePath.EndsWith("jpg", StringComparison.InvariantCultureIgnoreCase) || LogoFilePath.EndsWith("jpeg", StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        mimeType = "image/jpeg";
+                    }
+                    if (LogoFilePath.EndsWith("png", StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        mimeType = "image/png";
+                    }
+
```
KoenZomers, line 679: Should there not be an if none of the above scenario? Throw a warning or perhaps even an exception?

PR #2888, opened 2023-03-11
Added warning in the catch block

ChrisRo89:

## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

Added warning message for catch block related to MoveFile

### DIFF
--- a/src/Commands/Files/MoveFile.cs
+++ b/src/Commands/Files/MoveFile.cs
@@ -13,7 +13,7 @@ namespace PnP.PowerShell.Commands.Files
         private const string ParameterSet_SERVER = "Server Relative";
         private const string ParameterSet_SITE = "Site Relative";
         private const string ParameterSet_OTHERSITE = "Other Site Collection";
-
+        
         [Parameter(Mandatory = true)]
         [Alias("ServerRelativeUrl", "SiteRelativeUrl")]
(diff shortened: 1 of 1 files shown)


src/Commands/Files/MoveFile.cs
```diff
@@ -83,8 +83,9 @@ protected override void ExecuteCmdlet()
                         var folderServerRelativePath = folder.EnsureProperty(f => f.ServerRelativePath);
                         isFolder = folderServerRelativePath.DecodedUrl == ResourcePath.FromDecodedUrl(TargetUrl).DecodedUrl;
                     }
-                    catch
+                    catch (Exception ex)
                     {
+                        WriteWarning(ex.Message);
```
KoenZomers, line 88: I understand why you want to visualize this situation, not sure yet if warning is the best method for this as it may break automated builds that are using this today. Verbose might be a safer option. @gautamdsheth What's your take on this?

gautamdsheth: Yeah, `WriteVerbose` is the safer option for this exception, we don't want to break scripts.

ChrisRo89: All right thank you for you feedback - I'll change it and update the branch

PR #2934, opened 2023-03-24
Deprecates BlockEdit and BlockDeletion parameters of Set-PnPLabel. Closes #2932

### LINKED ISSUE #2932: [ENHANCEMENT] Deprecate -BlockEdit and -BlockDeletion parameters on Set-PnPLabel
Using the commandlet [Set-PnPLabel](https://pnp.github.io/powershell/cmdlets/Set-PnPLabel.html) you can apply a default retention label to lists and libraries. While doing that you can use the `-BlockEdit` and `-BlockDeletion` parameters to configure label settings. 

However, this has been a concern for some time (because these settings allow you to override the retention label settings in Microsoft Purview) That's the reason they have been disabled in the CSOM and REST API's. Using them is ignored by Microsoft, Microsoft will simply look up the correct values for both properties and use them in the label application. 

You can test this by applying a record retention label as a default to a list or libary and specify $false as a value for BlockEdit and BlockDelete. (A record-type label should have the value `true` on both these parameters)
After applying the label you can retrieve the applied label settings. You'll see your falsy values have been ignored.

```posh
Set-PnPLabel -List "TestLibrary" -Label "Test Recordlabel" -BlockDeletion:$false -BlockEdit:$false
Get-PnPlabel -List "TestLibrary"
```

These properties should therefore be deprecated and removed from the codebase.

martinlingstuyl:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [ ] Sample
- [x] Enhancement

## Related Issues? ##
Closes #2932

Deprecates `BlockEdit` and `BlockDeletion` parameters of `Set-PnPLabel`. The code has been updated to not use the deprecated parameters, but use the found retention label settings when calling the `ComplianceTag` method, even though that is not really necessary because using the properties have been cut off by Microsoft.

### DIFF
--- a/documentation/Set-PnPLabel.md
+++ b/documentation/Set-PnPLabel.md
@@ -38,17 +38,10 @@ Set-PnPLabel -List "Demo List" -Label "Project Documentation" -SyncToItems $true
 
 This sets an O365 label on the specified list or library and sets the label to all the items in the list and library as well.
 
-### EXAMPLE 3
-```powershell
-Set-PnPLabel -List "Demo List" -Label "Project Documentation" -BlockDelete $true -BlockEdit $true
-```
-
-This sets an O365 label on the specified list or library. Next, it also blocks the ability to either edit or delete the item. 
-
 ## PARAMETERS
 
 ### -BlockDeletion
-Block deletion of items in the library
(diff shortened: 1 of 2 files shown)


gautamdsheth:
Thanks @martinlingstuyl 😊🙏

PR #2940, opened 2023-03-26
Add PreviousNode parameter to Add-PnPNavigationNode

reusto:
## Type ##
- [x] Bug Fix
- [x] New Feature
- [ ] Sample

Added parameter `PreviousNode` to specify the navigation node after which the new navigation node will appear in the navigation node collection, which allows to add it to the correct position in the navigation instead of adding and then reordering afterwards.

Changed the type of the `Parent` parameter to NavigationPipeBind to allow a object of type NavigationNode as input in addition to just the ID.

Fixed the issue that target audiences weren't set when creating a node as a child of a parent node.

### DIFF
--- a/documentation/Add-PnPNavigationNode.md
+++ b/documentation/Add-PnPNavigationNode.md
@@ -14,8 +14,16 @@ Adds an item to a navigation element
 
 ## SYNTAX
 
+### Default
+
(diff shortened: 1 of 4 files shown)


KoenZomers:
Thanks @reusto!

PR #2942, opened 2023-03-27
Update Set-PnPFolderPermission.md

sumitkumar0608:
Fixed the Description of Set-PnPFolderPermission Cmdlets.

Added a Note that It breaks the inheritance of folders.


sumitkumar0608:
Fixed the Description of Set-PnPFolderPermission Cmdlets.

Added a Note that It breaks the inheritance of folders.

KoenZomers:
Thank you @sumitkumar0608 for contributing!

PR #2971, opened 2023-04-04
Adding option to write Graph output from `Invoke-PnPGraphMethod` to file

KoenZomers:
## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
N/A

Added `-OutFile` option to `Invoke-PnPGraphMethod` which allows the output of the request to be written to a file. This allows i.e. binary results such as calls to `https://graph.microsoft.com/v1.0/users/user@contoso.com/photo/`$value` to be written to a valid file instead of parsing it as a string first and then writing it to a file, which would not work for binary content.


gautamdsheth:
Thanks @KoenZomers !

Maybe we can also introduce something like Stream for folks to upload files using this cmdlet ? Something to think about ?

PR #2990, opened 2023-04-09
added new command Export-PnPPowerApp

siddharth-vaghasia:
## Type ##

- [x ] New Feature

## Related Issues? ##
NA

Added new command Export-PnPPowerApp which will export  a specified PowerApp as zip package, this command would be useful to automate the backup of the PowerApp/s or to automate the deployment in DevOps

Note - @gautamdsheth @KoenZomers  - I would need your advise/feedback specifically on one thing... so adding some comments in PR for your comment.


src/Commands/PowerPlatform/PowerApps/ExportPowerApp.cs
```diff
@@ -0,0 +1,181 @@
...
+                {
+                    displayName = PackageDisplayName,
+                    description = PackageDescription,
+                    creator = PackageCreatedBy,
+                    sourceEnvironment = PackageSourceEnvironment
+                },
+                resources = wrapper.Resources
+            };
+
+            var responseHeader = RestHelper.PostAsyncGetResponseHeader<string>(Connection.HttpClient, $"https://api.bap.microsoft.com/providers/Microsoft.BusinessAppPlatform/environments/{environmentName}/exportPackage?api-version=2016-11-01", AccessToken, payload: exportPostData).GetAwaiter().GetResult();
```
siddharth-vaghasia, line 104: @gautamdsheth  @KoenZomers 
To Export a PowerApp Package we need to make multiple API Calls, and this is 2nd call /exportpackage which returns us the location of another URL which will tell us when the download is ready but we have to keep of sending request.... and then when the status is successed it would return the url of zip file.  Now the location property from this API call comes in response Headers and not as a response content so I had to introduce new functions in RestHelper which will return us headers to read the response headers properties. 

So if you check in RestHelper, I have added two new functions...open to suggestions on the best way to handle this?

src/Commands/Utilities/REST/RestHelper.cs
```diff
@@ -633,6 +649,28 @@ private static async Task<string> SendMessageAsync(HttpClient httpClient, HttpRe
             }
         }
 
+        private static async Task<HttpResponseHeaders> SendMessageAsyncGetResponseHeader(HttpClient httpClient, HttpRequestMessage message)
```
siddharth-vaghasia, line 652: New method to make API Call but return Response Headers rather than content

src/Commands/Utilities/REST/RestHelper.cs
```diff
@@ -289,6 +289,22 @@ public static async Task<T> PostAsync<T>(HttpClient httpClient, string url, stri
             return default(T);
         }
 
+        public static async Task<HttpResponseHeaders> PostAsyncGetResponseHeader<T>(HttpClient httpClient, string url, string accessToken, object payload, bool camlCasePolicy = true, string accept = "application/json")
```
siddharth-vaghasia, line 292: Base method which is being called from actuall command.cs file.

gautamdsheth:
Can you please move the logic to a separate file or create one like `PowerAppsUtility` ? We can move the all these REST calls in there and keep the cmdlet clean and just return the link if available ? 

Also, can you please rename the method as `GetPackageLink` ? 

About the method names, am still thinking on how to reduce duplicate code, will get back to you on that

siddharth-vaghasia:
> Can you please move the logic to a separate file or create one like `PowerAppsUtility` ? We can move the all these REST calls in there and keep the cmdlet clean and just return the link if available ?
> 
> Also, can you please rename the method as `GetPackageLink` ?
> 
> About the method names, am still thinking on how to reduce duplicate code, will get back to you on that

Hi @gautamdsheth  - sorry was not able get to this sooner... I have updated as per review comments. please check now?

gautamdsheth:
Thanks @siddharth-vaghasia , merged this, amazing stuff ! Much appreciated 😊🙏
Sorry for late response.

siddharth-vaghasia:
hi @gautamdsheth - do let me know when it would be available for testing? I will test it once :)

KoenZomers:
@siddharth-vaghasia Should be in the current nightly build. Every night at 2.30 AM CET a nightly build gets created which contains the latest code thus far.

LucX4:
Hi @KoenZomers, @siddharth-vaghasia. Hope you are doing well!
I'm having trouble executing this cmdlet as it's not appearing in my list of available commands. Are there any prerequisites I might be missing?
Thank you in advance!

siddharth-vaghasia:
hi @LucX4 - can you please the PnP PowerShell version you are using?

LucX4:
Hi @siddharth-vaghasia,

It's 1.12.0:

image

LucX4:
@siddharth-vaghasia,
I wasn't using the latest version. I'm able to run the command now.
Thanks for your help.

LucasCXavier:
Hi @siddharth-vaghasia and @gautamdsheth, hope you're doing well!

I've been trying to run the Export-PnPPowerApp cmdlet for several hours but haven't had any success so far.
Even when authenticating via the .pfx certificate, I'm encountering a Forbidden error.

The error message says that the service principal doesn't have permission to access:

https://10.0.5.20:20273/providers/Microsoft.BusinessAppPlatform/environments/

The App Registration has already been added as System Administrator in the target environment, so it's unclear what's missing.

Has anyone faced something similar before or knows if there are additional permission steps required (perhaps at the Enterprise Application level or API permissions in Azure)?

Thanks,
Lucas

siddharth-vaghasia:
hi @LucasCXavier - Did you try running it with User's Permission to verify?...just connect pnp-online with any SharePoint site in interactive mode and the user should have access to targetted PowerApps as owner.

LucasCXavier:
Hi @siddharth-vaghasia, 
I got the following errror when trying to use Connect-PnPOnline -url https://tenant.sharepoint.com -interactive.
Connect-PnPOnline: Specified method is not supported.

Issue #1807, opened 2022-04-26
[BUG] Unable to create Team with Application ID and certificate


svermaak:
### Reporting an Issue or Missing Feature
Unable to create Team with Application ID and certificate

### Expected behavior 
Team created for M365 group

### Actual behavior
New-PnPTeamsTeam: No Team found with Group id: 6a************************ea
or no output
![image](https://user-images.githubusercontent.com/12259728/165238307-38e45df5-2855-4969-96b2-2c9654ed624e.png)
![image](https://user-images.githubusercontent.com/12259728/165238672-9353c052-917f-49af-8e77-33b78663893e.png)

### What is the version of the Cmdlet module you are running?
Latest

### Which operating system/environment are you running PnP PowerShell on?
- [x] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [x] Azure Functions
- [ ] Other : please specify

gautamdsheth:
Is it happening consistently ??  Can you try with a different groups or maybe on a different tenant maybe ?

svermaak:
I tried different owner and different group, same result

svermaak:
It fails here
![image](https://user-images.githubusercontent.com/12259728/165400090-f7f7fa20-9356-4480-980b-89a733941734.png)

![image](https://user-images.githubusercontent.com/12259728/165399992-7e9584a4-8673-4251-bab5-834dd01e9a3e.png)

veronicageek:
Closing as no response.

KoenZomers:
Should be fixed through https://github.com/pnp/powershell/pull/3351

PR #3351, opened 2023-08-17
Make New-PnPTeamsTeam work with Managed Identity

### LINKED ISSUE #2977: [BUG] - Unable to create teams using New-PnPTeamsTeam
Unable to create teams using New-PnPTeamsTeam. The pnp powershell version is 2.1.1
The Group is created, but the team never gets created.
When i user owner attribute , i get this error
```
New-PnPTeamsTeam: 
Line |
   5 |  $team = New-PnPTeamsTeam -DisplayName "write2" -Visibility Private  - …
     |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Bad Request (400): /me request is only valid with delegated authentication flow.
```

I am using managed identity in azure runbook. Following permissions are given to managed identity
image

When i create the team locally , it works

here is the code in runbook

```

$connection = Connect-pnponline -managedidentity -url "https://test-admin.sharepoint.com" -returnconnection;
Write-Output $connection
try
{
$team = New-PnPTeamsTeam -DisplayName "write3" -Visibility Private  -connection $connection -Owners "test@365909m365.onmicrosoft.com"
}
catch
{
    Write-Output $error
}


```

damsleth:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #2977, #1807 

* Adds team owners and members _after_ team creation, since calling the `/teams/{group.Id}/members/add` endpoint when there is no team fails with `New-PnPTeamsTeam: Not Found (404): No Team found with Group id: {groupId}`
* Inverts the boolean on [L296](https://github.com/pnp/powershell/blob/ccee58d064da543e40a4a7a872ad9574f8a108ed/src/Commands/Utilities/TeamsUtility.cs#L296) in TeamsUtility.cs, avoiding calls to the `/me` endpoint when an owner is already resolved. This is particularly useful when using a managed identity, since `/me` calls are only valid with a delegated auth flow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Fixed `New-PnPTeamsTeam` cmdlet, it will now throw error if it fails to teamify a Microsoft 365 group. [#3310](https://github.com/pnp/powershell/pull/3310)
 - Fixed `Connect-PnPOnline` cmdlet throwing host not reachable errors. [#3337](https://github.com/pnp/powershell/pull/3337)
 - Fixed `Set-PnPTerm` cmdlet throwing object reference error when only the term Id is specified. [#3341](https://github.com/pnp/powershell/pull/3341)
+- Fixed `New-PnPTeamsTeam` cmdlet throwing an error when specifying members [#3351](https://github.com/pnp/powershell/pull/3351)
+- Fixed `New-PnPTeamsTeam` cmdlet not working well with a managed identity [#3351](https://github.com/pnp/powershell/pull/3351)
 
 ### Changed
 
@@ -47,6 +49,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Contributors
 
+- Carl Joakim Damsleth [damsleth]
 - Rodel Pacurib [ryder-cayden]
 - [CatSchneider]
 - [msjennywu]
--- a/src/Commands/Utilities/TeamsUtility.cs
+++ b/src/Commands/Utilities/TeamsUtility.cs
@@ -195,33 +195,6 @@ namespace PnP.PowerShell.Commands.Utilities
             }
             if (group != null)
             {
-                // Construct a list of all owners and members to add
-                var teamOwnersAndMembers = new List<TeamChannelMember>();
-                if (owners != null && owners.Length > 0)
-                {
-                    foreach (var owner in owners)
-                    {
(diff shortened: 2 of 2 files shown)


wobba:
Is the call to `"v1.0/teams/{group.Id}/members/add` even needed? As all owners/members of a group should be default become members of the team? Or do two calls, where you add them as team members after the groups is initially created with owners/members.

KoenZomers:
Thanks @damsleth for your contribution! Appreciate it!

khha:
@damsleth @KoenZomers @wobba is it now possible to Teamify a non-group-based team site from an Azure function by a managed identity?

wobba:
@khha A site has to have an associated AAD group in order to be used in a Team. So you need to group'ify first still if you have a stand-alone site you want to use as a Team. So the description of this cmdlet stands as per https://pnp.github.io/powershell/cmdlets/New-PnPTeamsTeam.html

hajjkh:
@wobba but in the documentation I see "If the Microsoft 365 Group does not exist yet, it will create it first and then add a Microsoft Teams team to the group.". It looks like they build it by design to Groupify automatically before Teamify. Thats exactly where I am looking for because I would like to Groupify and Teamify.

KoenZomers:
> @damsleth @KoenZomers @wobba is it now possible to Teamify a non-group-based team site from an Azure function by a managed identity?

My motto is always seeing is believing :) Just give it a try and see if it works. One thing I stumbled upon myself is that in 2.5.0 it seems that something with regards to Managed Identity is broken. Still need to dig into that. Give it a try with 2.4.0 to play on the safe side of things.

wobba:
@hajjkh correct. The cmdlet does not take a SharePoint site as input, so the description is correct. Here is the technical ways you can get to a team.

1. Create Team from scratch. Creates a group behind the scenes first, then creates a team and associates it and create a SharePoint site and associate it
2. Create Team on a group. Creates a new Team and associate it with the AAD group
3. Create Team on a SharePoint site. You first need to create a group, and associate the site to the Group. Then create a team which is associated to the group as well.

This cmdlet cover scenario 1 and 2. Above, If someone wants to extend it to scenario 3, that is a possibility. Which is effectivly calling Add-PnPMicrosoft365GroupToSite first. 

As we're using powershell, just run the two commands in sequence :)

wobba:
May be that PnPMicrosoft365GroupToSite don't work with app permissions. Haven't checked.

hajjkh:
@wobba still not working in pnp version 2.4.0

PR #3396, opened 2023-09-04
Added Retention Label options to Set-PnPListItem

danielcecil:
## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
Adds the option to specify a user email and date when setting a Retention Label (ComplianceTag) to List Items.

Added two new parameters to the 'Single' param set in Set-PnPListItem to take a `LabelAppliedByUser` string and `LabelAppliedDate` DateTime object.

Forking in the logic that when the user email is specified then use a different CSOM call (SetComplianceTagWithMetaInfo) using provided values, and values from the ComplianceTag configuration (e.g. BlockDelete, BlockEdit, etc). AppliedDate defaults to current time when not specified as is the default behavior. 

Updated documentation with 1 new example and additional parameter descriptions.

### DIFF
--- a/documentation/Set-PnPListItem.md
+++ b/documentation/Set-PnPListItem.md
@@ -23,7 +23,7 @@ Updates a list item
 
 ```powershell
 Set-PnPListItem [-List <ListPipeBind>] -Identity <ListItemPipeBind> [-ContentType <ContentTypePipeBind>]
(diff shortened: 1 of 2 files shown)


danielcecil:
@gautamdsheth I know you recently tidied up some of the code around ComplianceTags so keen to hear your thoughts on the additional options to set a ComplianceTag with additional metadata.

PR #3397, opened 2023-09-04
[FEATURE] Added Retention Label options to Set-PnPListItem

danielcecil:
## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
Adds the option to specify a user email and date when setting a Retention Label (ComplianceTag) to List Items.

Added two new parameters to the 'Single' param set in Set-PnPListItem to take a `LabelAppliedByUser` string and `LabelAppliedDate` DateTime object.

Forking in the logic that when the user email is specified then use a different CSOM call (SetComplianceTagWithMetaInfo) using provided values, and values from the ComplianceTag configuration (e.g. BlockDelete, BlockEdit, etc). AppliedDate defaults to current time when not specified as is the default behavior. 

Updated documentation with 1 new example and additional parameter descriptions.

@gautamdsheth I know you recently tidied up some of the code around ComplianceTags so keen to hear your thoughts on the additional options to set a ComplianceTag with additional metadata.


gautamdsheth:
hi @danielcecil , code-wise it looks good.

I read this [post](https://learn.microsoft.com/en-us/sharepoint/dev/apis/csom-methods-for-applying-retention-labels), where it says: `It is strongly recommended to use SetComplianceTagOnBulkItems instead of these methods.` The CSOM method `SetComplianceTagWithMetaInfo` is also marked for deprecation, and they recommend using `SetComplianceTagOnBulkItems ` method.

So, don't think we should start adding a method marked for deprecation.

Also, checking with @martinlingstuyl , maybe he has some thoughts on this ?

gautamdsheth:
hi @danielcecil , will have to close this PR.
Unfortunately, the method has been marked as deprecated, so we shouldn't use it. 
Sorry about this, we can't merge it, hope you understand.

danielcecil:
Hi @gautamdsheth thank you for looking into it and your explanation. The docs weren't exactly clear but you and @martinlingstuyl have cleared that up and I agree not to add methods that may become deprecated in future. As I understand it this is the only way to add additional metadata to the application of the Compliance Tag, so it's an example of what I'm doing currently to meet this requirement.

martinlingstuyl:
In that case you probably should go back to the customer with the message: sorry we can't add extra metadata with when applying a label 😊

PR #3414, opened 2023-09-15
[New Feature] - Get-PnPTenantInfo - to get tenant info of any tenant

NishkalankBezawada:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

Gets information about any tenant

## SYNOPSIS
Gets information about any tenant

### By TenantId
```powershell
Get-PnPTenantInfo -TenantId <String>
```

### By DomainName
```powershell
Get-PnPTenantInfo -DomainName <String>
```

### Current Tenant
```powershell
Get-PnPTenantInfo
```

## DESCRIPTION

Gets information about any tenant. If no Domain name or Tenant id is specified, it returns the Tenant Info of the current tenant.

### EXAMPLE 1
```powershell
Get-PnPTenantInfo -TenantId "e65b162c-6f87-4eb1-a24e-1b37d3504663"

TenantId                             FederationBrandName DisplayName DefaultDomainName
--------                             ------------------- ----------- -----------------
e65b162c-6f87-4eb1-a24e-1b37d3504663                     contoso      contoso.onmicrosoft.com
```

Returns the tenant information of the specified TenantId as shown. A valid connection with Connect-PnPOnline is required either 

### EXAMPLE 2
```powershell
Get-PnPTenantInfo -DomainName "contoso.com"

TenantId                             FederationBrandName DisplayName DefaultDomainName
--------                             ------------------- ----------- -----------------
e65b162c-6f87-4eb1-a24e-1b37d3504663                     contoso      contoso.onmicrosoft.com
```

Returns the Tenant Information for the tenant contoso.sharepoint.com as shown. A valid connection with Connect-PnPOnline is required either 

### EXAMPLE 3
```powershell
Get-PnPTenantInfo

TenantId                             FederationBrandName DisplayName DefaultDomainName
--------                             ------------------- ----------- -----------------
e65b162c-6f87-4eb1-a24e-1b37d3504663                     contoso      contoso.onmicrosoft.com
```

Returns Tenant Information of the current tenant.

### EXAMPLE 4
```powershell
Get-PnPTenantInfo -CurrentTenant

TenantId                             FederationBrandName DisplayName DefaultDomainName
--------                             ------------------- ----------- -----------------
e65b162c-6f87-4eb1-a24e-1b37d3504663                     contoso      contoso.onmicrosoft.com
```

Returns Tenant Information of the current tenant.

### -TenantId
The id of the tenant to retrieve the information. You can use either TenantId or DomainName to fetch the Tenant information of any tenant. 

```yaml
Type: String
Parameter Sets: GETINFOBYTENANTID

Required: true
Position: Named
Default value: None
Accept pipeline input: False
```

### -DomainName
The Domain name of the tenant to lookup. You can use either TenantId or DomainName to fetch the Tenant information of any tenant. Use the full domain name as "contoso.onmicrosoft.com"

```yaml
Type: String
Parameter Sets: GETINFOBYTDOMAINNAME

Required: False
Position: Named
Default value: Production
Accept pipeline input: False
```

### -CurrentTenant
Gets the Tenant Information of the current tenant.

```yaml
Type: SwitchParameter
Parameter Sets: GETINFOOFCURRENTTENANT

Required: False
Position: Named
Default value: None
Accept pipeline input: False
```

## RELATED LINKS

[tenantRelationship: findTenantInformationByTenantId](https://learn.microsoft.com/en-us/graph/api/tenantrelationship-findtenantinformationbytenantid?view=graph-rest-1.0&tabs=http)

[tenantRelationship: findTenantInformationByDomainName](https://learn.microsoft.com/en-us/graph/api/tenantrelationship-findtenantinformationbydomainname?view=graph-rest-1.0&tabs=http)


gautamdsheth:
Hi @NishkalankBezawada - do you need this cmdlet ? We already have Get-PnPTenantId cmdlet which retrieves tenant Id value . 

What we usually do here is we add only cmdlets which we should need in your day to day life or which complete a story like the one you did PnPFlowOwner 😊

NishkalankBezawada:
> Hi @NishkalankBezawada - do you need this cmdlet ? We already have Get-PnPTenantId cmdlet which retrieves tenant Id value .
> 
> What we usually do here is we add only cmdlets which we should need in your day to day life or which complete a story like the one you did PnPFlowOwner
> 
> Adding more cmdlets means we will have to spend more time in maintaining it. So , need to add only those ones which we absolutely need.

Hello @gautamdsheth,

Agreed. But I did see some scope of enhancement in this command, and took an inspiration from [cli-m365-tenant-info-get](https://pnp.github.io/cli-microsoft365/cmd/tenant/info/info-get/). 

Should i close the PR then? 

Thanks,
Nishkalank Bezawada

KoenZomers:
Get-PnPTenantId just gets the tenant Id. We also have Get-PnPTenantInstance but that just gets the URLs and data locations of a tenant. This new cmdlet would allow to retrieve the default domain name. That could be useful and I don't believe we have anything as such today yet. I would say for the scenario where you don't provide any parameters, as I see in your example above, it would return an error. Could that be updated to show the info about the currently connected to tenant instead? That would fall more in line with how other cmdlets work today.

@gautamdsheth What do you think?

I would say we keep it

gautamdsheth:
@NishkalankBezawada / @KoenZomers - yeah ok lets keep it. Leaving some review comments.

src/Commands/Model/TenantInfo.cs
```diff
@@ -0,0 +1,11 @@
+﻿
+namespace PnP.PowerShell.Commands.Model
+{
+    public class TenantInfo
+    {
+        public string tenantId { get; set; }
```
gautamdsheth, line 6: All the names should be capitalized like this:

```
public string TenantId
```

src/Commands/Admin/GetTenantInfo.cs
```diff
@@ -0,0 +1,53 @@
...
+            var graphAccessToken = TokenHandler.GetAccessToken(this, $"https://{Connection.GraphEndPoint}/.default", Connection);
+            var requestUrl = BuildRequestUrl();
+
+            var results = RestHelper.GetAsync<Model.TenantInfo>(Connection.HttpClient, requestUrl, graphAccessToken).GetAwaiter().GetResult();
+            WriteObject(results, true);
+        }
+
+        private void EnsureValidInput()
+        {
+            if (string.IsNullOrEmpty(TenantId) && string.IsNullOrEmpty(DomainName))
```
gautamdsheth, line 32: You should use ParameterSets for this instead of this check, please check other cmdlets for example.

gautamdsheth (changes requested):

src/Commands/Admin/GetTenantInfo.cs
NishkalankBezawada: Hey @gautamdsheth 

Suggested changes are done accordingly.

Thanks,
Nishkalank Bezawada

NishkalankBezawada:
> Get-PnPTenantId just gets the tenant Id. We also have Get-PnPTenantInstance but that just gets the URLs and data locations of a tenant. This new cmdlet would allow to retrieve the default domain name. That could be useful and I don't believe we have anything as such today yet. I would say for the scenario where you don't provide any parameters, as I see in your example above, it would return an error. Could that be updated to show the info about the currently connected to tenant instead? That would fall more in line with how other cmdlets work today.
> 
> @gautamdsheth What do you think?
> 
> I would say we keep it

Hello @KoenZomers 

As suggested in this conversation, I have implemented 

```Get-PnPTenantInfo``` which will return the current tenant information

Thanks,
Nishkalank

KoenZomers:
Thanks @NishkalankBezawada , nice addition!

NishkalankBezawada:
> Thanks @NishkalankBezawada , nice addition!

Thanks @KoenZomers For your support.

PR #3426, opened 2023-09-20
Added dynamic membership option to New-PnPMicrosoft365Group

KoenZomers:
## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
N/A

Added the option to create a Microsoft 365 Group which will make use of dynamic membership instead of assigned membership only


src/Commands/Microsoft365Groups/NewMicrosoft365Group.cs
```diff
@@ -115,8 +133,16 @@ protected override void ExecuteCmdlet()
                     GroupTypes = new string[] { "Unified" },
                     PreferredDataLocation = PreferredDataLocation,
                     PreferredLanguage = PreferredLanguage,
+                    MembershipRule = DynamicMembershipRule
```
gautamdsheth, line 136: What happens if the `DynamicMembershipRule` is empty here ? Does it work as it works currently ?

gautamdsheth (changes requested):

src/Commands/Microsoft365Groups/NewMicrosoft365Group.cs
KoenZomers: Correct. At first I also had a SwitchParameter to indicate it needed to be a dynamic membership, but since the DynamicMembershipRule is required, I figured that could just as well already function as a trigger to indicate it needs to be dynamic instead of assigned.

PR #3459, opened 2023-10-02
Add new command get-pnpretentionlabel

reshmee011:
## Type ##
- [ ] New Feature

New command to get all Purview retention labels within tenant. The intent is to add more Purview retention label related commands to add, update, delete and get retention label by id.


src/Commands/Model/Graph/Purview/RetentionDuration.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        /// <summary>
+        /// The type of the data.
+        /// </summary>
+        [JsonPropertyName("@odata.type")]
+        public string ODataType { get; set; } = "#microsoft.graph.security.retentionDurationInDays";
+        /// <summary>
+        /// Number of days.
+        /// </summary>
+        public int days { get; set; }
```
gautamdsheth, line 15: Can you capitalize this to `Days` ?

src/Commands/Purview/GetRetentionLabel.cs
```diff
@@ -0,0 +1,25 @@
...
+using PnP.PowerShell.Commands.Utilities.REST;
+using System;
+using System.Collections.Generic;
+using System.Management.Automation;
+
+namespace PnP.PowerShell.Commands.Purview
+{
+    [Cmdlet(VerbsCommon.Get, "PnPRetentionLabel")]
+    [OutputType(typeof(IEnumerable<Model.Graph.Purview.RetentionLabel>))]
+    [OutputType(typeof(Model.Graph.Purview.RetentionLabel))]
```
gautamdsheth, line 12: I guess we can remove this since it will be a collection ?

gautamdsheth (changes requested):

src/Commands/Purview/GetRetentionLabel.cs
reshmee011: @gautamdsheth : planning to add parameter Id to retrieve retention label by id , would that affect the function if one record is returned ? ..shall I add the parameter Id in the current PR?  I have kept the initial implementation to get retention labels simple as was not sure what I did met PnP PowerShell standards.

src/Commands/Model/Graph/Purview/RetentionDuration.cs
reshmee011: I will capitalize Days.

src/Commands/Purview/GetRetentionLabel.cs
gautamdsheth: Have a question  - do you see the need for this cmdlet ? As in, would be using it in frequently in near future ? 

Really don't want to add more cmdlets, if not needed , since we have quite a few. Would rather focus on bug fixes and code quality. Gets difficult to maintain. Also, since this is beta , there's a good chance MS will break it.

reshmee011: PnP PowerShell has cmdlets to apply and get retention labels at site/list level 
Set-PnPLabel -List "<list_identifier>" -Label "Retain 10 years"
It would be great to have ability to manage retention labels within PnP PowerShell. I understand the concern of it being in still in beta and MS breaking it in the future. 
I would use it if it is within PnP PowerShell but can't tell about general usage by end users.

gautamdsheth: Ok fair point 😊
Yeah, would be good to have Identity parameter as well in this one.

src/Commands/Purview/GetRetentionLabel.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class GetAvailableRetentionLabel : PnPGraphCmdlet
```
gautamdsheth, line 13: Would also need `RecordsManagement.Read.All` as minimum permissions ? Can you please add that ? You can check other graph cmdlets where we have added different permissions.

If any issue, let me know :)

gautamdsheth (changes requested):

src/Commands/Purview/GetRetentionLabel.cs
gautamdsheth: This is great, thanks !!

Also, sorry, I wasn't clear. I meant add this attribute
`[RequiredMinimalApiPermissions("RecordsManagement.Read.All )]`

You can check `Add-PnPAzureADGroupMember` cmdlet for example 😊

reshmee011: Oops, misunderstood the requirement. Thanks for handling it. Thanks very much for merging it. I will add the other operations, to add, delete and update the retention label.

gautamdsheth:
Thanks a lot @reshmee011 , merged it !! 
Will be available from tomorrow onwards in nightly builds , much appreciated 🙏

reshmee011:
Thanks @gautamdsheth : I will work on the operations to add, set and delete retention labels next.

PR #3487, opened 2023-10-12
Adding `Get-PnPPowerPlatformConnector`

siddharth-vaghasia:
## Type ##
[x] New Feature

Added new command to Get PowerPlatform Solutions. This command would help to get the PowerPlatform solution for the given enviorment.


src/Commands/PowerPlatform/Environment/GetPowerPlatformConnectors.cs
```diff
@@ -0,0 +1,65 @@
...
+using PnP.PowerShell.Commands.Base;
+using PnP.PowerShell.Commands.Utilities.REST;
+using System;
+using System.Management.Automation;
+using System.Linq;
+using PnP.PowerShell.Commands.Base.PipeBinds;
+using System.Security.Cryptography.X509Certificates;
+
+namespace PnP.PowerShell.Commands.PowerPlatform.Environment
+{
```
gautamdsheth, line 11: It needs to be singular `PnPPowerPlatformConnector`

src/Commands/PowerPlatform/Environment/GetPowerPlatformConnectors.cs
```diff
@@ -0,0 +1,65 @@
...
+                WriteVerbose($"Using default environment as retrieved '{environmentName}'");
+            }
+
+            if (ParameterSpecified(nameof(Identity)))
+            {
+                var appName = Identity.GetName();
+
+                WriteVerbose($"Retrieving specific Custom Connector with the provided name '{appName}' within the environment '{environmentName}'");
+
+                var result = GraphHelper.GetAsync<Model.PowerPlatform.Environment.PowerPlatformConnector>(Connection, $"https://api.powerapps.com/providers/Microsoft.PowerApps{(AsAdmin ? "/scopes/admin/environments/" + environmentName : "")}/apis/{appName}?api-version=2016-11-01&$filter=environment eq '{environmentName}' and isCustomApi eq 'True'", AccessToken).GetAwaiter().GetResult();
```
gautamdsheth, line 52: Is the API url generic ? I guess it would be different for different clouds like GCC or GCC high ? Can you investigate it , if possible ? 

We have a somewhat similar thing for Graph, because Graph in commercial cloud is `graph.microsoft.com` whereas in GCC it would be `graph.microsoft.us`

src/Commands/PowerPlatform/Environment/GetPowerPlatformConnectors.cs
```diff
@@ -0,0 +1,65 @@
...
+                WriteObject(result, false);
+            }
+            else
+            {
+                WriteVerbose($"Retrieving all Connectors within environment '{environmentName}'");
+
+                var connectors = GraphHelper.GetResultCollectionAsync<Model.PowerPlatform.Environment.PowerPlatformConnector>(Connection, $"https://api.powerapps.com/providers/Microsoft.PowerApps/apis?api-version=2016-11-01&$filter=environment eq '{environmentName}' and isCustomApi eq 'True'", AccessToken).GetAwaiter().GetResult();
```
gautamdsheth, line 59: Same check here as well for GCC clouds

src/Commands/PowerPlatform/Environment/GetPowerPlatformSolution.cs
```diff
@@ -0,0 +1,65 @@
...
+
+                var requestUrl = dynamicsScopeUrl + "/api/data/v9.0/solutions?$filter=isvisible eq true and friendlyname eq '" + solutionName + "'&$expand=publisherid&api-version=9.1";
+                var solution = GraphHelper.GetResultCollectionAsync<Model.PowerPlatform.Environment.Solution.PowerPlatformSolution>(Connection, requestUrl, accessTokenForGettingSolutions).GetAwaiter().GetResult();
+                WriteObject(solution, false);
+            }
+            else
+            {
+                WriteVerbose($"Retrieving all Solutions within environment '{environmentName}'");
+                var requestUrl = dynamicsScopeUrl + "/api/data/v9.0/solutions?$filter=isvisible eq true&$expand=publisherid($select=friendlyname)&api-version=9.1";
+                //https://orge92ba4cd.crm8.dynamics.com/api/data/v9.0/solutions?%24expand=publisherid&%24filter=(isvisible%20eq%20true)&%24orderby=createdon%20desc
```
gautamdsheth, line 59: Can you please remove this comment ?

src/Commands/PowerPlatform/Environment/GetPowerPlatformSolution.cs
```diff
@@ -0,0 +1,65 @@
...
+                dynamicsScopeUrl = environments.FirstOrDefault(e => e.Properties.IsDefault.HasValue && e.Properties.IsDefault == true)?.Properties.LinkedEnvironmentMetadata.InstanceApiUrl;
+                if (string.IsNullOrEmpty(environmentName))
+                {
+                    throw new Exception($"No default environment found, please pass in a specific environment name using the {nameof(Environment)} parameter");
+                }
+
+                WriteVerbose($"Using default environment as retrieved '{environmentName}'");
+            }
+
+            string accessTokenForGettingSolutions = TokenHandler.GetAccessTokenforPowerPlatformSolutions(this, Connection, dynamicsScopeUrl);
```
gautamdsheth, line 43: Can we use `RequiredMinimalApiPermissions` attribute instead of this ? Please check some graph cmdlets where we use it.

src/Commands/PowerPlatform/Environment/GetPowerPlatformSolution.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                environmentName = environments.FirstOrDefault(e => e.Properties.IsDefault.HasValue && e.Properties.IsDefault == true)?.Name;
+                dynamicsScopeUrl = environments.FirstOrDefault(e => e.Properties.IsDefault.HasValue && e.Properties.IsDefault == true)?.Properties.LinkedEnvironmentMetadata.InstanceApiUrl;
+                if (string.IsNullOrEmpty(environmentName))
+                {
+                    throw new Exception($"No default environment found, please pass in a specific environment name using the {nameof(Environment)} parameter");
+                }
+
+                WriteVerbose($"Using default environment as retrieved '{environmentName}'");
+            }
```
gautamdsheth, line 41: Can you move this code to a utility ? 
It could be something like:

```
var env = Utility.GetEnvironment
```

Move the logic there and keep the cmdlet code itself smaller

src/Commands/PowerPlatform/Environment/GetPowerPlatformConnectors.cs
```diff
@@ -0,0 +1,65 @@
...
+using System;
+using System.Management.Automation;
+using System.Linq;
+using PnP.PowerShell.Commands.Base.PipeBinds;
+using System.Security.Cryptography.X509Certificates;
+
+namespace PnP.PowerShell.Commands.PowerPlatform.Environment
+{
+    [Cmdlet(VerbsCommon.Get, "PnPPowerPlatformConnectors")]
+    public class GetPowerPlatformConnectors : PnPAzureManagementApiCmdlet
```
gautamdsheth, line 13: Would require this I think:

```
[RequiredMinimalApiPermissions("https://management.azure.com/.default")]
```

src/Commands/PowerPlatform/Environment/GetPowerPlatformSolution.cs
```diff
@@ -0,0 +1,65 @@
...
+using PnP.PowerShell.Commands.Utilities.REST;
+using System;
+using System.Management.Automation;
+using System.Linq;
+using PnP.PowerShell.Commands.Base.PipeBinds;
+
+namespace PnP.PowerShell.Commands.PowerPlatform.Environment
+{
+    [Cmdlet(VerbsCommon.Get, "PnPPowerPlatformSolution")]
+    public class GetPowerPlatformSolution: PnPAzureManagementApiCmdlet
```
gautamdsheth, line 11: Same here:

```
Would require this I think:

```
[RequiredMinimalApiPermissions("https://management.azure.com/.default")]
```
```

gautamdsheth (changes requested):

src/Commands/PowerPlatform/Environment/GetPowerPlatformSolution.cs
siddharth-vaghasia: This token that we are getting would have a dynamic URL based on the Power Platform environment....so token has to be get for url like "https://orge12323.crm8.dynamics.com" and this will change - so this would not worked with RequiredMinimalPermissions?  Let me know if you think there is any other way to get token

src/Commands/PowerPlatform/Environment/GetPowerPlatformSolution.cs
siddharth-vaghasia: For PowerPlatform related commands I don't think it would require...

src/Commands/PowerPlatform/Environment/GetPowerPlatformSolution.cs
siddharth-vaghasia: yes...working on it

src/Commands/PowerPlatform/Environment/GetPowerPlatformConnectors.cs
siddharth-vaghasia: file commited by mistake..removed

siddharth-vaghasia:
hi @gautamdsheth - did you get a chance to review the updated comments?

KoenZomers:
If I may anothe review comment as well, could you rename the cmdlet to Get-PnPPowerPlatformCustomConnector? It only returns the custom connectors, not the out of the box ones. I find the current name confusing in this sense.

siddharth-vaghasia:
> If I may anothe review comment as well, could you rename the cmdlet to Get-PnPPowerPlatformCustomConnector? It only returns the custom connectors, not the out of the box ones. I find the current name confusing in this sense.

This is very good observation, changed the cmdlet to have custom word in it...

PR #3494, opened 2023-10-13
Fix retrieving error detail in Get-UPABulkImportStatus

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
None

This PR fixes the command `Get-UPABulkImportStatus -JobId "..." -IncludeErrorDetails` not showing the actual error detail.

Before the PR:
image

After the PR:
image

The issue was caused by how the relative path of the log folder was handled.

This actually triggered a code path previously never triggered, with an exception happening sometimes when trying to overwrite the ErrorMessage property.
Out of 11 import jobs with an error message, 3 had this issue. In these cases the previous generic message will be displayed instead, with the location of the logs on SPO.


gautamdsheth:
Thanks @jackpoz , have merged it , will be available in tomorrow's nightly builds

PR #3521, opened 2023-10-19
Allow to specify additional properties to include in Get-PnPSiteCollectionAdmin

jackpoz:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
No related issues

Allow to specify additional properties to include in `Get-PnPSiteCollectionAdmin`, for example `-Includes UserPrincipalName` , instead of having to rely on additional `Get-PnPProperty` calls like `Get-PnPProperty $spoAdmin UserPrincipalName` for each admin.

Considering the case of 1000 site collections and 3 site collection administrator users per site, this change allows to retrieve the UserPrincipalName property using 1000 requests in total instead of 4000 (1 per site plus 3 more per site).

The already available PnPWebRetrievalsCmdlet<T> class makes this feature easy to implement, with very few lines modified.


gautamdsheth:
Awesome addition @jackpoz , thank you so much !
Added it , will be available in tomorrow's nightly builds.

Issue #3569, opened 2023-11-14
[BUG] Get-PnPAzureADUser does not account for apostrophe's in UPN


danielcecil:
### Reporting an Issue or Missing Feature
When using Get-PnPAzureADUser any apostrophe's in the userPrincipalName are not escaped.

https://github.com/pnp/powershell/blob/0ac1939e7c871f67b32fee2eb795fe393d7161d3/src/Commands/Utilities/AzureAdUtility.cs#L86

### Expected behavior 
The Azure AD user is returned when passing in a userPrincipal name of a user (Member or Guest).

### Actual behavior 
`Get-PnPAzureADUser : Code: BadRequest
Message: Invalid filter clause: Syntax error at position 49 in 'userPrincipalName eq 'John.O'Smith_tailspintoys.com#EXT#@contoso.onmicrosoft.com''.`

### Steps to reproduce behavior
1. Create an Account with an apostrophe in the userPrincipalName
2. Run `Get-PnPAzureADUser -Identity "john.o'smith@contoso.com"`

### What is the version of the Cmdlet module you are running?
2.2.125
1.12.0

### Which operating system/environment are you running PnP PowerShell on?
- [X] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

KoenZomers:
PR has been merged. Will be in tomorrow's prerelease.

PR #3570, opened 2023-11-14
[FIX] #3569 Fixed fetching Azure AD User with apostrophe in UPN

danielcecil:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes (https://github.com/pnp/powershell/issues/3569)

Added a String.Replace function to the Identity parameter on GetAzureADUser.cs to replace single apostrophe's with doubles. e.g. `John.O'Smith@contoso.com` becomes `John.O''Smith@contoso.com` as a result.

This ensures the apostrophes are correctly escapes. Note that UrlEncode does not achieve this fix.


KoenZomers:
Thanks @danielcecil for your contribution! I have moved this fix to a lower level so two other cmdlets which ran into the same issue can also benefit from it.

danielcecil:
Nice work! Thanks for looking a bit deeper 👍

AlexSen:
Same issue with Get-PnPRetentionLabel
Even when I have site URL with apostrophe, I do connect by replacing it with %27, but Get-PnPRetentionLabel still decodes it and 

image


The expression "web/lists/getbytitle('/sites/XXXXXXXXXXXXBs_Write_Off's/Shared Documents')" is not valid.

PR #3599, opened 2023-11-24
Add Set-PnPComplianceTagOnBulkItems cmdlet

wilecoyotegenius:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

This PR adds a new cmdlet, Set-PnPComplianceTagOnBulkItems, which allows setting/clearing retention labels (aka compliance tags) from to one or more list items.


KoenZomers:
@wilecoyotegenius Nice one. Not a huge fan of the cmdlet name though. I see how this resembles the name it has been assigned in CSOM, but to me it would not be clear what it does by looking at the name. We already have a Set-PnPLabel which sets a label on a list. What if we add an alias on that one to be called Set-PnPSensitivityLabel so we can deprecate its current name in a future (major) version and merge your your code in that one. By using a paramset we can make it work for both the entire list as well as individual list items.

What do you think? Would that make sense?

wilecoyotegenius:
@KoenZomers This cmdlet does not set sensitivity labels but retention labels on items, so alias Set-PnPSensitivityLabel would be totally misleading. 
I am perfectly fine with adding a paramset to Set-PnPLabel.

wilecoyotegenius:
I can implement the paramset, if needed. Just let me know :)

KoenZomers:
Ah true.. there we go.. misleading naming :) What about renaming to Set-PnPRetentionLabel and adding it to that?

wilecoyotegenius:
Perfectly fine for me.

wilecoyotegenius:
One more idea - instead of combining cmdlets, we could leave them separate but rename current Get-PnPRetentionLabel to Get-PnPTenantRetentionLabel and then rename current Get-PnPLabel to Get-PnPRetentionLabel.
Leaving cmdlets separate might be reasonable due to the fact that currently code of Get-PnPRetentionLabel  and Get-PnPLabel is in different namespaces (Purview and InformationManagement respectively).

wilecoyotegenius:
@KoenZomers Shall we continue with this PR?

KoenZomers:
Apologies for the delay in my response. I like your proposal. I have updated the PR accordingly and will merge it. Thanks for your contribution!

Issue #3621, opened 2023-12-07
[BUG] Add-PnPNavigationNode OpenInNewTab not visible in UI


JuergenRB:
### Reporting an Issue or Missing Feature
Issue with `Add-PnPNavigationNode -OpenInNewTab`

### Expected behavior 
Link opens in a new tab and you can see the setting in the UI

### Actual behavior & Steps to reproduce behavior
If I set the property for a RootNode - it works and you can see the hook in the UI.

![image](https://github.com/pnp/powershell/assets/22403655/2c9338c1-9b8b-4d5d-a6df-1fbaa1a35a7c)
![image](https://github.com/pnp/powershell/assets/22403655/42bb839c-14c2-46b5-ac57-2e926c73c812)

If I set the property for a ChildNode with the Id or with the ParentNode as parent parameter, I get a warning and it is not visible in the UI. However, the link opens in a new tab.
![image](https://github.com/pnp/powershell/assets/22403655/8deae4c9-c0ae-4f4f-8d04-fc6bac6eb2c2)
![image](https://github.com/pnp/powershell/assets/22403655/dacbb9f6-accf-4e91-a764-8fc1a0a1bbc2)

If I look at the MenuState, I can see differences in the values / attributes for OpenInNewTab:

RootNode:
```XML
<d:Key>2028</d:Key>
<d:Nodes m:type="Collection(SP.MenuNode)"/>
<d:NodeType m:type="Edm.Int32">0</d:NodeType>
<d:OpenInNewWindow m:type="Edm.Boolean">true</d:OpenInNewWindow>
<d:SimpleUrl>https://github.com/pnp/powershell</d:SimpleUrl>
<d:Title>RootNode</d:Title>
```

SubNode:
```XML
<d:Key>2031</d:Key>
<d:Nodes m:type="Collection(SP.MenuNode)"/>
<d:NodeType m:type="Edm.Int32">0</d:NodeType>
<d:OpenInNewWindow m:null="true"/>
<d:SimpleUrl>https://github.com/pnp/powershell</d:SimpleUrl>
<d:Title>SubNode_ById</d:Title>
```

### What is the version of the Cmdlet module you are running?
2.2.0

### Which operating system/environment are you running PnP PowerShell on?
Windows

gautamdsheth:
Closing as fixed by PR.

PR #3625, opened 2023-12-07
Fix: #3621 - AddNavigationNode - Search in sub nodes for id

JuergenRB:
## Type ##
- [x] Bug Fix

## Related Issues? ##
Issue #3621

Within the AddNavigationNode command:

NavigationNode property Nodes now also contains NavigationNodes instead of objects. 
To find a node by ID, not only the first list but also all child nodes are searched.


gautamdsheth:
@JuergenRB - is this ready ? Have you tested the code ?

JuergenRB:
@gautamdsheth - Yes, the change is finished and tested

gautamdsheth:
Thanks @JuergenRB , good catch, much appreciated the fix for the issue !!

PR #3638, opened 2023-12-17
Created a new Cmdlet for Mergeing Terms and a documentation page.

joHKwi:
## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
This was an issue for some tasked i solved - so maybe it will be useful for someone else as well. 

Takes one term and merges it into the another term.


src/Commands/Taxonomy/MergeTerms.cs
```diff
@@ -0,0 +1,54 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Management.Automation;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace PnP.PowerShell.Commands.Taxonomy
+{
+    [Cmdlet(VerbsData.Merge, "PnPTerms")]
```
gautamdsheth, line 13: It should be `PnPTerm` . We cannot allow plural as per PowerShell convention.

Can you also update the same in docs ?

joHKwi: Done - in both pluss file names.

PR #3676, opened 2024-01-09
fix for get message replies includeDelete flag

### LINKED ISSUE #3532: [BUG] Get-PnPTeamsChannelMessageReply and -IncludeDeleted flag bug
### Reporting an Issue 

Get-PnPTeamsChannelMessageReply command is returning only deleted replies if we don't specify -IncludeDeleted flag 

`Get-PnPTeamsChannelMessageReply -Team MyTestTeam -Channel "My Channel" -Message 1653089769293 -IncludeDeleted`

### Actual behavior 

`Get-PnPTeamsChannelMessageReply -Team CLITeamDemo -Channel "climicrosoft365" -Message 1698072147830`

If we are not specifying -IncludeDeleted option it is only returns deleted messages replies, below screenshot 

Reference message screenshot (if you see here there are 4 replies and only 1 is deleted)
image

Output
image

### Expected behavior 
if we don't specify -IncludeDeleted option it should return all the active messages/which are not deleted

Run the command with below option

Get-PnPTeamsChannelMessageReply -Team CLITeamDemo -Channel "climicrosoft365" -Message 1698072147830

PnP.Powershell 2.0.53 (nightly build)

- [x ] Windows

siddharth-vaghasia:
## Type ##
- [X ] Bug Fix

## Related Issues? 
Fixes #3532 

Fix to only return Deleted message replies when -IncludeDeleted flag is set as true. Before the fix - without specifying this parameter also it was returning deleted messages replies.

### DIFF
--- a/src/Commands/Utilities/TeamsUtility.cs
+++ b/src/Commands/Utilities/TeamsUtility.cs
@@ -705,7 +705,7 @@ namespace PnP.PowerShell.Commands.Utilities
         {
             var replies = await GraphHelper.GetResultCollectionAsync<TeamChannelMessageReply>(connection, $"v1.0/teams/{groupId}/channels/{channelId}/messages/{messageId}/replies", accessToken);
 
-            return includeDeleted ? replies.ToList() : replies.Where(r => r.DeletedDateTime.HasValue).ToList();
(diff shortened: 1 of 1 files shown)


gautamdsheth:
Thanks @siddharth-vaghasia, good catch , merged it !!!

PR #3677, opened 2024-01-09
fix for changing base URL for Get-Flow

### LINKED ISSUE #3640: [BUG] Get-PnPFlow: Bad Request (400)
Hi. I tried running the `Get-PnPFlow` command today, and it looks like it is calling a deprecated API. It spits out `Get-PnPFlow: Bad Request (400): Calling Power Automate through ARM is no longer supported. Please use "https://api.flow.microsoft.com/` instead.

I pulled the address from [GetFlow.cs](https://github.com/pnp/powershell/blob/master/src/Commands/PowerPlatform/PowerAutomate/GetFlow.cs) and tried it in Postman, and I received the same result.

Ironically, it appears [the API](https://learn.microsoft.com/en-us/business-applications-release-notes/October18/microsoft-flow/flow-management-connector-sdk-and-apis) it refers to is being deprecated, so [this](https://learn.microsoft.com/en-us/power-automate/manage-flows-with-code?tabs=webapi) may be relevant.

siddharth-vaghasia:
## Type ##
- [X ] Bug Fix

## Related Issues? ##
Fixes #3640 

Fix to change the base url as Command was returning "Bad Request (400): Calling Power Automate through ARM is no longer supported. " 
Reference issue/conversation in CLI repo - https://github.com/pnp/cli-microsoft365/issues/5715

We would have to update all the base URL for PowerAutomate commands.... I am working on same...will create new issue/PR for other command as well.

### DIFF
--- a/src/Commands/PowerPlatform/PowerAutomate/GetFlow.cs
+++ b/src/Commands/PowerPlatform/PowerAutomate/GetFlow.cs
@@ -57,7 +57,7 @@ namespace PnP.PowerShell.Commands.PowerPlatform.PowerAutomate
 
                 WriteVerbose($"Retrieving specific Power Automate Flow with the provided name '{flowName}' within the environment '{environmentName}'");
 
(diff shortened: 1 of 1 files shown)


gautamdsheth:
Thanks @siddharth-vaghasia , merged it , much appreciated !!

PR #3715, opened 2024-01-29
fix: some properties of ExtractConfiguration are not used

nils-a:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
fixes https://github.com/pnp/pnpframework/issues/953

Some properties of `ProvisioningTemplateCreationInformation` were always set, regardless of them being explicitly set in the Cmdlet, or not.

Thus, those properties could not be set by using an `ExtractConfiguration`.

### DIFF
--- a/src/Commands/Provisioning/Site/GetSiteTemplate.cs
+++ b/src/Commands/Provisioning/Site/GetSiteTemplate.cs
@@ -213,14 +213,26 @@ namespace PnP.PowerShell.Commands.Provisioning.Site
                 creationInformation.PersistBrandingFiles = PersistBrandingFiles;
             }
 #pragma warning restore 618
-            creationInformation.PersistPublishingFiles = PersistPublishingFiles;
-            creationInformation.IncludeNativePublishingFiles = IncludeNativePublishingFiles;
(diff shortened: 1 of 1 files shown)


KoenZomers:
Thanks @nils-a for your contribution. Can you please edit this PR and allow us to update your PR? We need it to include the changelog entry.

nils-a:
@KoenZomers This is a trap I always fall into. While editing of this PR is allowed, it does not work since you have no permissions on my fork. This is a "feature" of GitHub, when forks are created in an organization...

I have given you write permissions on my fork, so you should now be able to push into it.

KoenZomers:
Thanks @nils-a. Not sure what it is, but it still cannot interact with your PR branch. Typically only just leaving the default checkbox checked when creating your PR should be enough. No need to share entire repositories:

image

For now let's not waste any more time on this. I have taken your changes and created a new PR https://github.com/pnp/powershell/pull/3741 through which I will merge them.

Thanks for your efforts!

KoenZomers:
Merged through https://github.com/pnp/powershell/pull/3741

PR #3762, opened 2024-02-16
update get-pnpfolder to remove reference to Get-PnPFolderFile and Get-PnPFolderFolder 

reshmee011:
## Type ##
- [ ] Bug Fix

Remove the following from docs as cmdlets Get-PnPFolderFile and Get-PnPFolderFolder does not exist

```
Use [Get-PnPFolderFile](Get-PnPFolderFile.md) to retrieve only files and [Get-PnPFolderFolder](Get-PnPFolderFolder.md) to retrieve only folders allowing additional properties of the returned items to be requested.
```


gautamdsheth:
Thanks @reshmee011 , good catch. We had renamed these cmdlets before, so made the necessary changes.

reshmee011:
@gautamdsheth : I can access the page https://pnp.github.io/powershell/cmdlets/Get-PnPFolderFolder.html but can't find it in the solution, does it need to be removed?

PR #3784, opened 2024-02-23
Update set-pnpfield with no crawl to prevent the field from being returned in search results

reshmee011:
## Type ##
- [ ] New Feature

Add no crawl property to set-pnpfield Update set-pnpfield to prevent the field from being returned in search results


reshmee011:
@gautamdsheth : I have updated the docs. Forgot to push the changes for the docs earlier..

gautamdsheth:
@reshmee011 - do we really need this ? I mean , we can easily do it via something like this. 
Works for almost all properties.

```
Set-PnPField -Identity "Test" -Values @{NoCrawl=$true}
```

reshmee011:
@gautamdsheth , updating NoCrawl property is failing using method  -values @{NoCrawl=$true}  as well the switchparameter -nocrawl.

image

Something changed to prevent the update to the NoCrawl at column level. I will raise it with Microsoft and wait for an answer. 

Happy for the PR to be closed and to use the -values option to update the property.

gautamdsheth:
Ok, sure. Will keep it open, let me know what MS responds to this , we can merge it later then.
Thanks :)

reshmee011:
It seems to be able to set the field as non crawlable, the "SharePoint Server Publishing Infrastructure" has to be activated which is not a recommended solution , hence closing the PR.

reshmee011:
It seems to be able to set the field as non crawlable, the "SharePoint Server Publishing Infrastructure" has to be activated which is not a recommended solution , hence closing the PR.

PR #3788, opened 2024-02-24
[New Parameters to Existing Commandlet] - Added new parameters to Like and Unlike the page

NishkalankBezawada:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

Added new parameters as `-Like` which will like and unlike the page respectively. 

## Example for Like
```powershell
Set-PnPPage -Identity "MyPage.aspx" -Like
```
Likes the page

### Example for Unlike
```powershell
Set-PnPPage -Identity "MyPage.aspx" -Like:$false
```
Unlikes the page

### -Like
Likes the page

```yaml
Type: SwitchParameter
Parameter Sets: (All)

Required: False
Position: Named
Default value: None
Accept pipeline input: False
```

Thanks,
Nishkalank Bezawada


src/Commands/Pages/SetPage.cs
```diff
@@ -64,6 +64,12 @@ public class SetPage : PnPWebCmdlet, IDynamicParameters
         [Parameter(Mandatory = false)]
         public bool ShowPublishDate;
 
+        [Parameter(Mandatory = false)]
+        public SwitchParameter Like = false;
+
+        [Parameter(Mandatory = false)]
+        public SwitchParameter UnLike = false;
```
gautamdsheth, line 71: I think we can use a single parameter for this ? Just keep the parameter `Like` and in the code we check if its present. If the value of it is true , we like the page , if false, we unlike.

So end users will have to do something like 

```
Set-PNPPage "123.aspx" -Like
```

or 

```
Set-PnPPage "123.aspx" -Like:$false
```

gautamdsheth (changes requested):

src/Commands/Pages/SetPage.cs
NishkalankBezawada: Hello @gautamdsheth, 

Thanks for the suggestion. I have changed accordingly now, Kindly review. 

Regards,
Nishkalank Bezawada

PR #3891, opened 2024-04-12
Fix NullDereferenceException in Get-PnPUserProfileProperty

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
No related issues

Fix NullDereferenceException happening in Get-PnPUserProfileProperty when the user profile doesn't exist, showing a better error message.
This happens because ServerObjectIsNull is set to true when UPS doesn't find an account.

Repro steps:
```powershell
Get-PnPUserProfileProperty -Account "test@test.com"
Get-PnPUserProfileProperty -Account "test1@gpsgit.onmicrosoft.com" -Properties ABC
Get-PnPUserProfileProperty -Account "test1@gpsgit.onmicrosoft.com" -Properties Title
```

Before the PR:
image

After the PR:
image

With an existing account:
image

When passing properties that don't exist, like "ABC", no exception is returned due to how that particular case is handled, using a different PeopleManager method. The behavior is the same before and after this PR.


gautamdsheth:
Thanks @jackpoz , merged it !!

PR #3937, opened 2024-05-05
Allow to pass empty values to Set-PnPTenantCdnPolicy

### LINKED ISSUE #3935: [BUG] Set-PnPTenantCdnPolicy doesn't accept null policy values
On a newly created tenant I was applying several policies through Microsoft365DSC and noticed that due to a bug there, which I still didn't report or created a PR for, it applied a Tenant CDN policy with *ExcludeRestrictedSiteClassifications* set to what I had previously applied on *IncludeFileExtensions* which of course is wrong, but to my surprise I couldn't set ExcludeRestrictedSiteClassifications back to empty with *Set-PnPTenantCdnPolicy* as it was the default in the tenant since *PolicyValue* is a string and can't be null. The same applies for *IncludeFileExtensions*, these 2 parameters by default are null in the tenant and the cmdlet should be able to restore this but can't. FYI, I was able to resolve this by running the following M365 CLI command on a PS prompt:

```powershell
m365 spo cdn policy set --cdnType Public --policy ExcludeRestrictedSiteClassifications --value @()
```

Also note that even though that command solved my issue it also has a problem since applying it with *$null* instead of *@()* the value it stores in the tenant is *TRUE* which is incorrect, but that is a separate issue that should be checked with M365 CLI's team.

### Expected behavior 
*Set-PnPTenantCdnPolicy* should accept for both *IncludeFileExtensions* and *ExcludeRestrictedSiteClassifications* a null policy value, both parameters should have AllowEmptyString in https://github.com/pnp/powershell/blob/dev/src/Commands/Admin/SetTenantCdnPolicy.cs

### Actual behavior 
*Set-PnPTenantCdnPolicy* fails if policy value parameter is null since it's a string and mandatory

```powershell
Set-PnPTenantCdnPolicy -CdnType "Public" `
    -PolicyType 'IncludeFileExtensions' `
    -PolicyValue $null
```

or

```powershell
Set-PnPTenantCdnPolicy -CdnType "Public" `
    -PolicyType 'ExcludeRestrictedSiteClassifications' `
    -PolicyValue $null
```

1.12.0, the reason for the old version is due to a compatibility issue with M365DSC and therefore cannot be upgraded, but the problem is present on latest version of https://github.com/pnp/powershell/blob/dev/src/Commands/Admin/SetTenantCdnPolicy.cs 

- [X] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #3935

Change Set-PnPTenantCdnPolicy to allow PolicyValue parameter to be an empty string or $null, while still being mandatory. Users will still need to specify -PolicyValue parameter.
This is the same behavior as when calling `Tenant.SetTenantCdnPolicy(..)` in CSOM.

Example with IncludeFileExtensions:
image
image

Example with ExcludeRestrictedSiteClassifications:
image

Example with ExcludeIfNoScriptDisabled:
image

`$null` is replaced with an empty string to avoid a runtime exception. This was the error happening without the replacement:
null error


gautamdsheth:
Thanks @jackpoz , awesome stuff , merged it !!

PR #3992, opened 2024-06-01
Fix New-PnPTeamsTeam sometimes "Conflict (409): Team already exists"

### LINKED ISSUE #3964: [BUG] New-PnPTeamsTeam sporadically returns "Conflict (409): Team already exists"
We are using PnP Powershell 2.3.0 for provisioning of new Teams within one of our Azure Function Apps. For the New-PnPTeamsTeam cmdlet we have seen the strange behavior that this is sporadically (maybe in 1 out of 20 runs) failing with the error "Conflict (409): Team already exists" even though I can guarantee that this Team does not exist at this point of time. I have also checked what error occurs if I create another Team with the same mail nickname which will give me a more specific error than the one that we are seeing (using the same mailnickname twice will cause "New-PnPTeamsTeam: Bad Request (400): Another object with the same value for property mailNickname already exists.")

After checking the failed creation I can see that the Team got created nevertheless so the error also is misleading. It looks to me that during the creation of the Team there might be some retries or else happening within the Framework which are causing this error.

Is anyone else seeing the same behavior?

This is the line of code that fails with Conflict 409 sporadically

`New-PnPTeamsTeam -DisplayName $TeamDisplayName -Description $TeamDescription -MailNickName $MailNickName -Visibility $Visibility -Owners $TeamOwners -Members $TeamMembers -AllowAddRemoveApps $false -AllowChannelMentions $true -AllowCreateUpdateChannels $false -AllowCreateUpdateRemoveConnectors $false -AllowCreateUpdateRemoveTabs $false -AllowCustomMemes $true -AllowDeleteChannels $false -AllowGiphy $true -AllowGuestCreateUpdateChannels $false -AllowGuestDeleteChannels $false -AllowOwnerDeleteMessages $false -AllowStickersAndMemes $true -AllowTeamMentions $true -AllowUserDeleteMessages $false -AllowUserEditMessages $false -GiphyContentRating "Moderate" -ShowInTeamsSearchAndSuggestions $true`

2.3.0

- [ ] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [X] Azure Functions
- [ ] Other : please specify

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #3964

Handle "Conflict (409): Team already exists" error as a case of "a previous teamify call succeeded" in New-PnPTeamsTeam when teamifying the newly created M365 group.

As explained at https://github.com/pnp/powershell/issues/3964#issuecomment-2131624168 , the Graph call https://learn.microsoft.com/en-us/graph/api/team-put-teams?view=graph-rest-1.0&tabs=http returns 404 when calling it right after creating the group but sometimes actually teamifies the group, causing an error "Conflict (409): Team already exists" to any later call. This PR handles that case of 429 as success.

To test this I reproduced the case pausing New-PnPTeamsTeam execution with a breakpoint, teamifying the group with Graph Explorer and resuming PowerShell execution. This returned the team correctly:
image

This might not be the most foolproof solution but it's a quite practical one and I used a similar approach in different provisioning solutions that interacted directly with Graph.


gautamdsheth:
Thanks @jackpoz , merged it !

PR #3993, opened 2024-06-01
new cmdlet for Add-PnPTenantRestrictedSearchAllowedList

reshmee011:
## Type ##
- [ ] New Feature

Files to support new cmdlet for Add-PnPTenantRestrictedSearchAllowedList

documentation\Add-PnPTenantRestrictedSearchAllowedList.md
src\Commands\Admin\AddTenantRestrictedSearchAllowedList.cs


KoenZomers:
Thanks @reshmee011 . It feels like a paramset would be a better fit here as opposed to using Dynamic Parameters. Would you agree or is there a specific reason you chose to go with this option?

reshmee011:
@KoenZomers : Hello, it makes more sense to use paramset to be consistent and easier to maintain in the future. I will refactor the code.

reshmee011:
@KoenZomers : I have updated the cmdlet to use parameter set, thanks very much for the review.

KoenZomers:
Thanks @reshmee011 !

PR #4029, opened 2024-06-18
[FEATURE] Added the Folder parameter to Add-PnPDocumentSet

### LINKED ISSUE #4022: [FEATURE] Add-PnpDocumentSet requires -Folder parameter
The file
https://github.com/pnp/powershell/blob/dev/src/Commands/DocumentSets/AddDocumentSet.cs

Uses the root folder of the document library and does not accept a folder as a parameter.

Add a folder parameter like the add-pnplistitem command.

Probably need to manually use the SP client api awkwardly.

**Additional context**
I'm trying to make document sets within folders in a document library.

danielcecil:

## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #4022

Adds a String -Folder parameter to the command, similar to Add-PnPListItem. Folder is created if needed.

Includes updated documentation too


src/Commands/DocumentSets/AddDocumentSet.cs
```diff
@@ -19,6 +19,9 @@ public class AddDocumentSet : PnPWebCmdlet
         [ValidateNotNullOrEmpty]
         public string Name;
 
+        [Parameter(Mandatory = false)]
+        public string Folder;
```
gautamdsheth, line 23: Can you please modify this to use `FolderPipeBind` ? You can check & copy over the implementation that we have in `Add-PnPFolder` cmdlet, would make things easier and consistent for users

gautamdsheth (changes requested):
Need to use the implementation that we have in `Add-PnPFolder` for folder parameter

src/Commands/DocumentSets/AddDocumentSet.cs
danielcecil: Thanks @gautamdsheth I can update this, however I used the logic in `Add-PnPListItem` which will create the folder if it doesn't exist. It also uses the relative path from the `List.RootFolder` e.g. `"Folder A/Folder B"` as in this command we are specifying the List already which makes sense rather than having to repeat the List name/path in the `FolderPipeBand` e.g. `"Shared Documents/Folder A/Folder B"`

gautamdsheth: @danielcecil - will merge it and make some changes later. The advantage of pipebind is that we can easily use it with `Get-PnPFolder` or related cmdlets. If you notice the code, it does create the folder if it doesn't exist in that cmdlet.

PR #4072, opened 2024-07-04
Adding notice to source code of `New-SPOContainerType`

KoenZomers:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [ ] Sample
- [X] Other

## Related Issues? ##
N/A

Following up on internal discussion

### DIFF
--- a/src/Commands/Admin/NewContainerType.cs
+++ b/src/Commands/Admin/NewContainerType.cs
@@ -7,9 +7,10 @@ using System.Management.Automation;
 namespace PnP.PowerShell.Commands.Admin
 {
     [Cmdlet(VerbsCommon.New, "PnPContainerType")]
-    public class NewContainerType : PnPAdminCmdlet,IDynamicParameters
+    public class NewContainerType : PnPAdminCmdlet, IDynamicParameters
     {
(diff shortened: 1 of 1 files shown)


src/Commands/Admin/NewContainerType.cs
```diff
@@ -29,6 +30,15 @@ public object GetDynamicParameters()
             return null;        
         }
 
+        //
+        // NOTICE: The SharePoint API being used in this code is of temporary nature.
```
javieralvarezchiang, line 34: My main concern is that SPContainerTypeProperties is a construct that lives only in the cmdlets as part of a two-step action executed at that layer (that actually interacts with two different systems) and will not be reused when the Graph APIs are created. Instead, those two steps will be kept as two distinct steps.

By keeping AzureSubscriptionId, Region and ResourceGroup as part of your contract, you'd be forcing your users to add information not needed at that point, and probably missing things that can be added.

On the other hand, you can always say that you'll keep this interface for backwards compatibility, even if those three are not used. In fact, I'm curious to know if this works with non-trial containers (trial containers do not require that second step, but standard do, and that step is not executed if you interact with the Tenant object)

Issue #4082, opened 2024-07-09
[BUG] New-PnpSite with Managed Identity gives "Object reference not set to an instance of an object. " but still creates the Site


CGrossNteam:
### Reporting an Issue or Missing Feature
I'm using the "New-PnPSite" cmdlet inside of a Runbook, which is used with a Managed Identity.

The Managed Idetity has the following Permissions:

- Microsoft Graph [Directory.ReadWrite.All]
- Microsoft Graph [Team.Create]
- Microsoft Graph [Group.Create]
- Microsoft Graph [TeamMember.ReadWrite.Al]
- Office 365 SharePoint Online [Sites.FullControl.All]

I connect to the SharePoint Admin Site the following way:

`$ConnectionAdminCenter   = Connect-PnPOnline -Url "https://tenant-admin.sharepoint.com" -ManagedIdentity -ReturnConnection`

And then use the "New-PnpSite" cmdlet this way:

`New-PnPSite -Type TeamSite -Alias "Site-Alias" -Title "Site-Name" -Description "Description" -Owners "owner@tenant.com" -Lcid 1031 -Wait -Connection $ConnectionAdminCenter `

When I try to create a "TeamSiteWithoutMicrosoft365Group" it works without issue.

### Expected behavior 
The Site will be created with the parameters given. After that the runbook will continue with what comes afterwards.

### Actual behavior 
The Site is created and after about 10 minutes the Runbook shows the Status "Error" and in the "Exceptions" Tab I see:

`Object reference not set to an instance of an object. (Object reference not set to an instance of an object.)`

### What is the version of the Cmdlet module you are running?
I'm running Version 2.5.12-nightly of the PnP.Module.

### Which operating system/environment are you running PnP PowerShell on?
- [x] Windows
- [ ] Linux
- [ ] MacOS
- [x] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

jackpoz:
Could you try passing `-ErrorAction:Stop` as additional parameter and then logging more exception details with `Get-PnPException | Select-Object *` ?

CGrossNteam:
This is the Exception I'm getting:

`Message          : Object reference not set to an instance of an object.
Stacktrace       :    at PnP.PowerShell.Commands.NewSite.ExecuteCmdlet() in 
                   D:\a\powershell\powershell\src\Commands\Admin\NewSite.cs:line 145
[...]

PR #4087, opened 2024-07-10
Handle cases where site creation fails in New-PnPSite

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Improves #4082

Handle cases where site creation fails in New-PnPSite by throwing a clearer exception.
The previous code would try to dereference `returnedContext` even when null, throwing a `Object reference not set to an instance of an object.` exception. This happens because `Framework.Sites.SiteCollection.Create()` returns null in some cases, like if the site url was not retrieved after enough tries.

I'm not sure if the message should be changed. At this point of the callstack we have no idea what went wrong during the creation and if the site was fully created and configured or not. It might be that there was an issue only when retrieving the SPO ClientContext but everything else worked fine.
In any case, anything should be better than an unhandled null exception 😃

### DIFF
--- a/src/Commands/Admin/NewSite.cs
+++ b/src/Commands/Admin/NewSite.cs
@@ -130,7 +130,8 @@ namespace PnP.PowerShell.Commands
 
                 if (ClientContext.GetContextSettings()?.Type != Framework.Utilities.Context.ClientContextType.SharePointACSAppOnly)
                 {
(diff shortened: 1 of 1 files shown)


gautamdsheth:
hi @jackpoz , am going to close this PR , have refactored the teamsite creation code quite a lot which fixes quite a lot of issues that we have had related to some errors users have faced. It should cover most of the scenarios.

Issue #2043, opened 2022-06-28
Clear-PnPDefaultColumnValues does nothing?


jeroenbrussich:
### Discussed in https://github.com/pnp/powershell/discussions/2034

<div type='discussions-op-text'>

<sup>Originally posted by **jeroenbrussich** June 25, 2022</sup>
I have a Documentlibrary on which I want to remove some default Column Values.
Idealiter, I only would want to remove some default Values on some folders, but right now I'm desperate enough to nuke all default Values on all Folders and just start over :)

Chances are I don't quite understand how the command is run, so any help would be welcome.

```
$sharepointSite = "mySharepoint.online/sites/SITE"
Connect-PnPOnline -Url $sharePointSite -ClientId $clientID -Tenant $tenant -Thumbprint $thumbPrint -ErrorAction Stop

# $folder is a folder in `documents`  (List ID= 'C45604CF-73A7-451E-98A0-F0791D8FC3C7)

#  FOLDER = "100008"

Clear-PnPDefaultColumnValues -List 'documents' -Field werkterrein -Folder '100008' 
Clear-PnPDefaultColumnValues -List 'C45604CF-73A7-451E-98A0-F0791D8FC3C7' -Field werkterrein -Folder '100008' 

#   FOLDER = 100008

Clear-PnPDefaultColumnValues -List 'documents' -Field werkterrein -Folder 100008 
Clear-PnPDefaultColumnValues -List 'C45604CF-73A7-451E-98A0-F0791D8FC3C7' -Field werkterrein -Folder 100008 

#   NO FOLDER - RESET FIELD ON TOPFOLDER

Clear-PnPDefaultColumnValues -List 'documents'
Clear-PnPDefaultColumnValues -List 'C45604CF-73A7-451E-98A0-F0791D8FC3C7' 

```

My parameters seem to be correct, When I change $sharepointSite to only "mySharepoint.online", I get an error.
However, when I fiddle with -List, or -Field, things are becoming interesting.

The command runs for an equal amount of time, whether I include -Field or not.
What's more, when I -List a non-existent List (Let's say: 'C45604CF-73A7-451E-98A0', or DOCS'), the command keeps running the same amount of time. I get no errors. So it looks like It's timing out?

Using the parameter -verbose does nothing.

So I'm stuck, don't quite know what's going on/wrong and can't debug.

HELP :)
</div>

CallumCrowley:
[...]

PR #4088, opened 2024-07-10
Improve Set-PnPDefaultColumnValues and Clear-PnPDefaultColumnValues

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Related to #2043

Improve Set-PnPDefaultColumnValues and Clear-PnPDefaultColumnValues
- Add "/" prefix to Clear-PnPDefaultColumnValues -Folder parameter when missing
- Throw an exception if the field is not found
- Remove unneeded null checks as exceptions are thrown instead

## Test cases
1. Passing `-Field ""` was simply doing nothing, now it shows an error
Before:
Before - Empty field
After:
After - Field not found

1. Passing `-Folder` to `Clear-PnPDefaultColumnValues` without a "/" as prefix was doing nothing, now it adds the prefix as needed
Before (notice how Clear does nothing and the default field value is still returned):
Before - No slash
After:
After - No slash
With "/" specified:
After - With slash

1. No changes when passing `$null` or `""` to `-List` or when passing `$null` to `-Field`:
After - Same validation errors

I recommend viewing the diff with "Hide whitespace" enabled due to the formatting changes:
image


gautamdsheth:
Thanks @jackpoz , merged it !

PR #4167, opened 2024-08-17
Possiblity to overwrite default X509KeyStorageFlags in Connect-PnPOnline

PeterRevsbech:

## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #X, partially fixes #Y, mentioned in #Z, etc.

The optional parameter X509KeyStorageFlags is added to the Connect-PnPOnline cmdlet. This is relevant when authenticating with a certficate in order to specify where and how to import the private key of the X.509 passed certificate.
The X509KeyStorageFlags variable is an enum supporting bitwise operations as described here:
https://learn.microsoft.com/dotnet/api/system.security.cryptography.x509certificates.x509keystorageflags

This variable is passed in the X509Certificate2 constructor, which the cmdlet uses for certificate authentication.
https://learn.microsoft.com/dotnet/api/system.security.cryptography.x509certificates.x509certificate2.-ctor

Previously, it was not possible to overwrite the default X509KeyStorageFlags used in the constructor. 
This is problematic in cases that the default key storage flags did not support.

**Example:** In order to use Connect-PnPOnline from a PowerShell Azure Function App with a self-signed PKCS certificate that is fetched from a key vault, you would have to use the cmdlet like so:
`Connect-PnPOnline -Url $SiteUrl -ClientId $clientId -CertificateBase64Encoded $certBase64 -Tenant $spTenant`
The cmdlet then currently creates the certificate in the file `src/Commands/Base/ConnectOnline.cs` with this line
`var certificate = new X509Certificate2(certificateBytes, CertificatePassword);`

Leaving out the optional parameter X509KeyStorageFlags, effectively forcing it to be the default flags.
In an Azure Function app, that has no user store for private keys, the constructor would throw a WindowsCryptographicException error like this:
`ERROR: The system cannot find the file specified. ERROR: The system cannot find the file specified.`

This is fixed by setting the X509KeyStorageFlags variable to the flag "MachineKeySet", allowing the private key to be stored in the local computer store. I.e. adding this parameter to the cmdlet would fix the issue.
`Connect-PnPOnline -Url $SiteUrl -ClientId $clientId -CertificateBase64Encoded $certBase64 -X509KeyStorageFlags [System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::MachineKeySet -Tenant $spTenant`


gautamdsheth:
Merged via https://github.com/pnp/powershell/pull/4324 , will be available in tomorrow's nightly builds. 

Thanks a lot for investigating this and submitting a PR , much appreciated 😊🙏

PR #4213, opened 2024-08-30
Fix exception in Get-PnPDiagnostics

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
No related issues

Fix exception in Get-PnPDiagnostics happening when there is no Correlation Id.

This is the exception from Visual Studio:
image

After the PR the command works correctly:
image


gautamdsheth:
@jackpoz - any idea why the contains check doesn't work ? We do check the message for correlationId , maybe we can improve that ?

jackpoz:
The contains works and `Data` contains an element with key "CorrelationId", it's just that this element is `null` and calling `ToString()` throws the exception.

gautamdsheth:
Thanks @jackpoz , makes sense, merged it !!

PR #4242, opened 2024-09-06
Add `Invoke-PnPAdminRestMethod` cmdlet

autinerd:
## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

This adds the Cmdlet `Invoke-PnPAdminRestMethod` to invoke a REST method in the admin context.

An example is for retrieving the SharePoint Storage Quotas of the tenant at `/_api/StorageQuotas()?api-version=1.3.2` of the tenant SharePoint Admin URL.

This code is a clone of the `Invoke-PnPSPRestMethod`, but using the AdminContext.

### DIFF
--- a/documentation/Invoke-PnPAdminRestMethod.md
+++ b/documentation/Invoke-PnPAdminRestMethod.md
@@ -1,0 +1,140 @@
+---
+Module Name: PnP.PowerShell
+schema: 2.0.0
+applicable: SharePoint Online
+online version: https://pnp.github.io/powershell/cmdlets/Invoke-PnPAdminRestMethod.html
+external help file: PnP.PowerShell.dll-Help.xml
+title: Invoke-PnPAdminRestMethod
+---
+ 
+# Invoke-PnPAdminRestMethod
+
+## SYNOPSIS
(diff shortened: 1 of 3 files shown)


autinerd:
Oh, you are right! I just tested it and it works!

In this case then, the problem is that https://pnp.github.io/powershell/cmdlets/Connect-PnPOnline.html or other documentation doesn't mention at all that you can just use the admin site in the -Url parameter and then do API calls againt it with the normal `Invoke-PnPSPRestMethod`. So then my use case is fulfilled, thanks.

For the parameter, this was my first idea as well, but I couldn't find an easy way to get the `AdminContext` in the normal `Invoke-PnPSPRestMethod` cmdlet (but I'm not really familiar with the codebase, so maybe it's just me who couldn't figure it out).

gautamdsheth:
@autinerd - would like to update the documentation and update it with an example ? Docs are there in the `documentation` folder.

gautamdsheth:
Closing this , do send a PR if you want to add an example for connecting to admin site.
Thanks again for taking the time to submit the PR.

PR #4274, opened 2024-09-11
Fix exception in Set-PnPMicrosoft365GroupSettings

### LINKED ISSUE #4263: [BUG] Set-PnPMicrosoft365GroupSettings is returning type conversion error
When calling `Set-PnPMicrosoft365GroupSettings -Identity $groupSettingId -Values @{"AllowToAddGuests"="true"} -Group $groupId` I get the below error:
**Set-PnPMicrosoft365GroupSettings: Cannot implicitly convert type 'void' to 'object'**.

I have verified both the $groupSettingId and $groupId values and they exist.
In fact, I could create the group setting on the same $groupId using `New-PnPMicrosoft365GroupSettings` just a couple of minutes earlier, but with `-Values @{"AllowToAddGuests"="true"}`.

Besides using the id's for the group and group setting, I have also tried using the entire group object as well as the group setting object with no difference.

Workaround right now is to use `Remove-PnPMicrosoft365GroupSettings -Identity $groupSetting.Id -Group $groupId -Connection $conn` which works fine, after which the original `New-PnPMicrosoft365GroupSettings` can be run but with the new Values hash table.

### Expected behavior 
That the group setting to be updated.

### Actual behavior 
Receiving the error: Set-PnPMicrosoft365GroupSettings: Cannot implicitly convert type 'void' to 'object'

Run cmdlet

2.12.0

- [x] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #4263

Fix exception in Set-PnPMicrosoft365GroupSettings due to old async C# code no longer needed. Originally the methods were async C# methods returning Task, now they simply return void.
Very strange C# behavior that this code even passed the build and threw a runtime exception.

Before the PR:
image

After the PR (somehow I couldn't find a M365 group with working settings, but at least it doesn't throw anymore a C# runtime error but a Graph exception):
image


gautamdsheth:
Thanks @jackpoz , much appreciated. Yeah, this seems to be the missing thing.

thaHoly7:
Tried with 2.99.111-nightly but produces same error

Issue #4310, opened 2024-09-18
[BUG] set-pnpplannertask not updating records


skaufman-bbins:
PNP.SharePoint module v2.12.0 on PowerShell 7.4.5

cmdlet:  Set-PNPPlannerTask

### Reporting an Issue or Missing Feature
Set-PNPPlannerTask is not assigning users

### Expected behavior 
Before the switch (PNP.SharePoint v2.9) to using the new security model, the set-pnpplannertask would update the assignedto on tasks

### Actual behavior 
Upgrading to PNP.Sharepoint 2.12.0 & going through the process of registering the new application, the set-pnpplannertask is no longer assigning users to tasks

### Steps to reproduce behavior
#> $clientid = "xxxxx-xxx-xx-xx-guid"
#> $tenant = "tenantdnsname.onmicrosoft.com"
#> connect-pnponline -Url https://xxxxx.sharepoint.com -Interactive -clientid $ClientID -tenant $tenant -Verbose
VERBOSE: PnP PowerShell Cmdlets (2.12.0)
VERBOSE: Connecting using Interactive login
VERBOSE: Using ClientID 5b37e74f-01ab-4baa-a968-c18cfba99d21
VERBOSE: Connected
---Logon with my GA account or service account.  Both accounts have permissions to the Planner & the underlying M365 group.

#> get-pnpconnection

ConnectionMethod                           : Credentials
ConnectionType                             : O365
InitializationType                         : Unknown
Scopes                                     :
PSCredential                               :
ClientId                                   : xxxxxxx-xxx-x-x-xxxxxxxxx
ClientSecret                               :
ApplicationInsights                        : PnP.PowerShell.ALC.ApplicationInsights
Url                                        : https:///xxxxx.sharepoint.com/
TenantAdminUrl                             :
Certificate                                :
DeleteCertificateFromCacheOnDisconnect     : False
Context                                    : PnP.Framework.PnPClientContext
Tenant                                     :
UserAssignedManagedIdentityObjectId        :
UserAssignedManagedIdentityClientId        :
UserAssignedManagedIdentityAzureResourceId :
AzureEnvironment                           : Production

#> $groupID = "xxxxxx-xxx-x-xxx-xx-x"  #the group ID where the Planner resides
#>  $planID = "xxxxx-x-xxx-xxx"  #The GUID of the Planner in the group
[...]

PR #4327, opened 2024-09-21
Handle errors in Graph batch requests

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Improves https://github.com/pnp/powershell/issues/4310 by displaying an error when there are missing permissions

Handle errors in Graph batch requests by throwing an exception at the first found error.
Example:
image

Before the PR, no error was displayed or handled by the code:
image

I saw that GraphException and GraphError classes are used when throwing Graph related exceptions but I'm unsure what would be the best design in this case. Should it throw an exception with just the first error or should it throw an AggregateException with all the errors ?
This is how AggregateException would look in PowerShell:
image
using this sample code:
```csharp
throw new AggregateException(
    new Exception("first ex"),
    new Exception("second ex"));
```


gautamdsheth:
Am wondering how it should work in case some requests succeed and some fail.
Maybe we show all the errors like the Aggregate errors that you mentioned ?

jackpoz:
This is how it looks with the latest changes with 1 and 2 exceptions in the batch:
image

I added a little description text `$"{errors.Count} error(s) occurred in a Graph batch request"` to help users troubleshoot what went wrong and where, feel free to change the wording.

In this particular case, the request to get the id of my own account succeed, while getting the id of other accounts fails due to missing User.Read.All delegated Graph permissions.

I'm in no hurry to get this PR merged and I'm not sure which cmdlet rely on ignoring errors, it does save a lot of time though to see the exception message for those troubleshooting why some cmdlets are not working, especially due to the recent change of permissions in app registrations.

gautamdsheth:
Thanks @jackpoz , merged it. It is looks better now. Much better than no errors and users left wondering what went wrong.

The ErrorAction parameter if specified will handle this as well.

PR #4357, opened 2024-09-29
Updating verbose token permission validation - Replaced

KoenZomers:
## Type ##
- [ ] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
N/A

Updating the oauth permission validation shown when using -Verbose on cmdlets to support sovereign clouds and not to throw an exception when the permission check fails, but rather show a warning instead.

Sample:

Delegated login with verbose
image

App only login with verbose
image

App only login without verbose
image


KoenZomers:
Replaced by https://github.com/pnp/powershell/pull/4366

Issue #4380, opened 2024-10-04
[BUG]Get-PnPSearchCrawlLog throws access denied although accounts has enough permission


mikarin20:
### Reporting an Issue or Missing Feature
Please confirm what it is that your reporting

Running Get-PnPSearchCrawlLog but throwing error
Get-PnPSearchCrawlLog: Make sure you are granted access to the crawl log via the SharePoint search admin center at https://-admin.sharepoint.com/_layouts/15/searchadmin/crawllogreadpermission.aspx

my account has connection and permissions on crawllog but still facing issues to access crawl logs. What am i missing?

### Expected behavior 
Please describe what output you expect to see from the PnP PowerShell Cmdlets

see the logs from crawls

### Actual behavior 
Please describe what you see instead. Please provide samples of output or screenshots.

![image](https://github.com/user-attachments/assets/a9f46ea3-8016-4187-aed5-88a38f577d96)

### Steps to reproduce behavior
Please include complete script or code samples in-line or linked from [gists](https://gist.github.com/)
get-pnpsearchcrawllog

### What is the version of the Cmdlet module you are running?
(you can retrieve this by executing ```Get-Module -Name "PnP.PowerShell" -ListAvailable```)

2.12

### Which operating system/environment are you running PnP PowerShell on?
- [x ] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

jackpoz:
How did you authenticate ?

If I authenticate with -Interactive to the SPO Admin Center with an app registration with delegated permissions SharePoint AllSites.FullControl , I get the log as soon as I add my account to /_layouts/15/searchadmin/crawllogreadpermission.aspx :

![image](https://github.com/user-attachments/assets/bdac18b4-4b9d-4ed6-b517-b78a13289b8c)

mikarin20:
I authenticated with clientid and secret values via app register. It's connecting to spo fine. I'm able to retrieve a lot of information related file size, modified date, names etc but not this one.
I have also full permission like you.

Im not sure what could be missing

jackpoz:
Well, if you connect through clientid and secret, then that's Application permissions, not Delegated permissions, and /_layouts/15/searchadmin/crawllogreadpermission.aspx allows to specify only users, so Delegated permissions.

It might be that this particular API doesn't support application permissions but only delegated permissions, so with -Interactive or -DeviceLogin with an account.

PR #4391, opened 2024-10-05
Add warning to Get-PnPSearchCrawlLog with application permissions

jackpoz:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Related to #4380

Add a warning to Get-PnPSearchCrawlLog with application permissions as it seems supports only delegated permissions.

Using a certificate returns unauthorized even with Sites.FullControl.All , but to be 100% sure one would have to ask SPO product group to confirm that there is no support at all for application permissions to this API.

This is the error displayed with application permissions:
image


gautamdsheth:
Thanks @jackpoz , no way for us as such to reach PGI except for raising support ticket. 
Don't have capacity for this , hopefully, this warning will help users.

PR #4457, opened 2024-10-20
New cmdlet to update retention label on file

reshmee011:
## Type ##
- [ ] New Feature

Files related to new cmdlet to unlock a list item

-src\Commands\Lists\UnlockListItemRecord.cs
- documentation\Unlock-PnPListItemRecord.md


gautamdsheth:
hey @reshmee011 , instead of this CSOM based API request , let's use the graph based one:

https://learn.microsoft.com/en-us/graph/api/driveitem-lockorunlockrecord?view=graph-rest-1.0&tabs=http

Can you modify the PR to do that ?

reshmee011:
@gautamdsheth : Sure I can do that, I will make the changes soon.

reshmee011:
@gautamdsheth : Is there any existing function to retrieve the driveid of the list or library from Graph?
The graph endpoint requires the drive id, if there is already functionality for it, it will help.
https://learn.microsoft.com/en-us/graph/api/driveitem-lockorunlockrecord?view=graph-rest-1.0&tabs=http

gautamdsheth:
Hi @reshmee011 , we use PnP Core for this.

Check this :

https://github.com/pnp/powershell/blob/1b8af72aa866cd2898a1c728c56841be29a81317/src/Commands/Files/ConvertFile.cs#L67

Maybe we can use the VroomDriveId I think. Also, I think the value is same as list GUID value AFAIK, need to check , so list.ID or list.GUID might work

reshmee011:
@gautamdsheth : Thanks for the guidance.  I have not tested it yet, will `Connection.PnPContext.Web.GetFileByServerRelativeUrl` work for list item as well? Is there a method to just get List/library driveid within PnP Core?

reshmee011:
@gautamdsheth : I have used the REST API endpoint /_api/SP.CompliancePolicy.SPPolicyStoreProxy.UnlockRecordItem() which works for both a file and list item. The Graph endpoint works against a driveitem which from my understanding is applicable only to a file. I bet another cmdlet Unlock-PnPFile would work for the Graph endpoint https://learn.microsoft.com/en-us/graph/api/driveitem-lockorunlockrecord?view=graph-rest-1.0&tabs=http

gautamdsheth:
Hmm, need to check this a bit about how to fetch driveId via PnP Core. 
We want to avoid CSOM moving forward and use Graph where possible especially for new cmdlets. If it's not possible (list item) via Graph, lets not do it.

Will get back to you asap.

gautamdsheth:
@reshmee011 - we can fetch the Drive ID and Drive ItemID as below:

```
IFile fileToDownload = Connection.PnPContext.Web.GetFileByServerRelativeUrl(serverRelativeUrl);
fileToDownload.EnsureProperties(f => f.VroomDriveID, f => f.VroomItemID);

var itemId = fileToDownload.VroomItemID;
var driveId = fileToDownload.VroomDriveID;
```

We can then use these values in the Graph patch request

reshmee011:
@gautamdsheth :Thank you. Shall I also update the cmdlet to Unlock-PnPFileRecord to reflect that it will be applicable to files and not to list items.?

gautamdsheth:
Maybe lets call it `Update-PnPFileRetentionLabel` ?

reshmee011:
@gautamdsheth : There is another endpoint to update retention label https://learn.microsoft.com/en-us/graph/api/driveitem-setretentionlabel?view=graph-rest-1.0&tabs=http, I can implement is as Update-PnPFileRetentionLabel. 

For this one which refers , shall I keep as Unlock-PnPFileRecord with a switchparameter to lock the record? 
https://learn.microsoft.com/en-us/graph/api/driveitem-lockorunlockrecord?view=graph-rest-1.0&tabs=http

gautamdsheth:
Both of them are point to the same Graph endpoint with different payloads.
Would be great if we can combine both in a single cmdlet with different parameters like name , isRecoredLocked. What do you think ?

reshmee011:
Sounds good, I will have a go at it and reiterate with further changes.

reshmee011:
@gautamdsheth : sorry about the delay. I finally updated the PR to use the MS Graph and applies only to file retention label updates.

KoenZomers:
Thanks @reshmee011 !

Issue #4472, opened 2024-10-23
[BUG] With changes to M365 App Registration, we need detailed information on which permissions are required for which cmdlets


sparkle263:
### Reporting an Issue or Missing Feature
Several PNP Cmdlets are still failing after creating the Tenant-specific App registration.
Because our scripts were working prior to this change, and some are working after, we know it is likely due to some permissions that are still missing in this new registration.
However, when this happens, there is no specific information provided on what is missing (Error is generic).

I am also not finding any specific documentation anywhere - per cmdlet - of which permissions are required to run successfully. 
(This is needed for several/all cmdlets within this module, Request-PnPPersonalSite being one of them for today.)

Our account is being used Interactively (due to MFA), and has the full SharePoint Administrator Role (All sites in Tenant).

Or if there is any resource you can point me to where we can test to identify which permissions are missing / needed. 
i.e. documentation / web tool like they have for MS Graph etc.

### Expected behavior 
For this Cmdlet, perform the task as expected, and if not provide more information as to why / what permission(s) are missing.

### Actual behavior 
For this specific cmdlet (there are others), this is the error information returned:
ERROR: [ Request-PnPPersonalSite : Attempted to perform an unauthorized operation. ]
Error Exception Type Fullname: [ Microsoft.SharePoint.Client.ServerUnauthorizedAccessException ]

### Steps to reproduce behavior
For this specific cmdlet (there are others), this is the syntax used:
Request-PnPPersonalSite -UserEmails "user1@ourdomain.com" -NoWait -Connection $MyPnPConnection -Verbose

### What is the version of the Cmdlet module you are running?
We are using Module version 1.12.0 due to limitation with PS ISE / 5.1.
(We can make this work, but with the new app registration requirement, we need to know more information on the permissions needed)

### Which operating system/environment are you running PnP PowerShell on?
- [ X ] Windows (Server 2022)

alexsie48:
Could not agree more. We have started running into this problem with multiple cmdlets. We need more information to make these fixes.

gautamdsheth:
Closing this since we have a duplicate , https://github.com/pnp/powershell/issues/4353

Issue #4353, opened 2024-09-27
[DOCS] Add required permissions to each cmdlet


sympmarc:
With the recent changes to the app registration for the module, it would be great to have a consistent bit of info on each cmdlet showing which permissions are required for it to work. Some cmdlets have this already, like [Get-PnPAzureADAppSitePermission](https://pnp.github.io/powershell/cmdlets/Get-PnPAzureADAppSitePermission.html). Others, like [Get-PnPSite](https://pnp.github.io/powershell/cmdlets/Get-PnPSite.html) do not.

![image](https://github.com/user-attachments/assets/adf35179-b05e-48b4-90dd-efc7fe03487a)

![image](https://github.com/user-attachments/assets/d1a02e73-441f-45e4-987c-a95a498298a5)

If there's a way to generate a list of cmdlets which don't have this info, we could crowdsource the content.

Understanding what permissions are required to accomplish specific things can be problematic, as shown in issue #4351. While it might seem anyone running PnP.PowerShell should inherently know which permissions are required, I'm positive that's not the case.  It certainly isn't obvious to me most of the time, now that I'm having to think about it with the recent changes. In many cases, we have to ask someone else to consent to the permissions (totally reasonable) and it is important we know which permissions to ask for.

jackpoz:
It's worth to mention that cmdlet parameters affect which permissions are needed.

Take https://pnp.github.io/powershell/cmdlets/Set-PnPPlannerTask.html as example, with permissions "Microsoft Graph API: One of Tasks.ReadWrite, Tasks.ReadWrite.All, Group.ReadWrite.All". When using the "-AssignedTo" parameter, then "User.Read.All" is also needed, which led to https://github.com/pnp/powershell/issues/4310 .

Maybe these additional permissions should be listed in the details of each parameter ? Or maybe this information fits better on top in the existing "Required Permissions" but in a different way to show that, if you don't use all parameters, you don't need all permissions ?

veronicageek:
@sympmarc - That's a very valid ask, and we will work on that. Many cmdlets using Graph (i.e.: Teams, Planner, EntraID) have already good documentation around the Graph permissions but also appreciate what @jackpoz is saying. SharePoint cmdlets use mostly SPO API at this time (not Graph), so... lots of work in perspective. 

And of course, anyone can jump in! 😉

sympmarc:
[...]

PR #4473, opened 2024-10-23
Add warning to Get-PnPTenantInfo about required permissions

NishkalankBezawada:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #X, partially fixes #, mentioned in #4472 and #4353 , etc.

Add a warning to Get-PnPTenantInfo. Currently It just gives `403: Insuffiecient priviledges to complete the operation`. So with the change, It specifies the required permissions to be granted on Azure AD App to run this command. 

After the change, The command shows a warning as below, 

image

Thanks,
Nish


src/Commands/Admin/GetTenantInfo.cs
```diff
@@ -7,6 +8,7 @@
 namespace PnP.PowerShell.Commands.Admin
 {
     [Cmdlet(VerbsCommon.Get, "PnPTenantInfo")]
+    [RequiredApiApplicationPermissions("graph/CrossTenantInformation.ReadBasic.All")]
```
gautamdsheth, line 11: I think we should also add `RequiredApiDelegatedPermissions("graph/CrossTenantInformation.ReadBasic.All")` here since it works with delegated as well if available. What do you think ?

gautamdsheth (changes requested):

src/Commands/Admin/GetTenantInfo.cs
NishkalankBezawada: Agree with you, Changed the attribute to `RequiredApiDelegatedOrApplicationPermissions`

gautamdsheth:
Thanks @NishkalankBezawada , merged it !! 
Much appreciated .

Issue #36, opened 2020-11-28
[FEATURE] Add site Id to output of Get-PnPTenantSite


alexbelevantsev:
**Is your feature request related to a problem? Please describe.**
Get-PnPTenantSite doesn't return site Id.

**Describe the solution you'd like**
Get-PnPTenantSite returns a site Id.

**Describe alternatives you've considered**
Currently we use the following cmdlets to get the site Id:
```$tenant_site_url =  "https://contoso-admin.sharepoint.com"
Connect-PnPOnline -Url $tenant_site_url -Credentials $cred
$sites=Get-PnPTenantSite -Detailed
foreach($tenant_site in $sites){
    Connect-PnPOnline -Url $tenant_site.Url -Credentials $cred
    $site = Get-PnPSite -Includes ID
}
```
Unfortunately additional cmdlets takes much time to execute.

erwinvanhunen:
Hi @alexbelevantsev , 

Unfortunately the API we call behind the scenes does not return a site id. The cmdlet returns the objects as returned by that API. If we would include the site id, we basically would have to do the same as you're already doing in your script, which would mean we would have to create our own object to return (minor problem, but still it would add a task to our work where we would have to make sure that our object stays in sync with the one being returned from the APi, and make sure that all properties from the API object are available in our own object), but it will also add a serious performance hit in the cmdlet. Especially with tenants that have thousands of thousands of sites. There is also the challenge that it can be possible that a tenant admin has no access to a specific site collection and retrieving the properties will fail then for that site collection. 

I would propose to create an entry in user voice: https://sharepoint.uservoice.com/ so that the engineering team see the request and hopefully will add it to a future version of the APIs we use.

fthorild:
Hi!

Now, I am a bit rusty but I ended up in a situation where I needed to map a usage report only having SiteId and I wanted to solve it via pwsh but couldn't.

Instead I solved it with a console app but I do think what I did is pretty much the same thing that happens under the covers here? Or am I wrong?

In this file (src/Commands/Admin/GetTenantSite.cs) it looks like that but the SiteId is missing from the model in this file (src/Commands/Model/SPOSite.cs)

Just wanted to see if this after all is possible to fix and also maybe fairly simple? @erwinvanhunen 

![site-props](https://github.com/user-attachments/assets/39d3178f-6fbb-4779-af11-81d5b077b2bb)

KoenZomers:
Fixed through https://github.com/pnp/powershell/pull/4527

PR #4527, opened 2024-11-07
Added SiteId to the model since it was missing

fthorild:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
SiteId was not included in the SPOSite model and therefore not possible to get back even though it is there under the covers. In my case I needed to merge some stuff with the SP usage reports which only has SiteId as a key. This is mentioned way back in #36   

Two lines of code in the SPOSite object to include the SiteId


KoenZomers:
Thanks for your contribution @fthorild, much appreciated!

PR #4578, opened 2024-11-24
Update documentation to use Microsoft.PowerShell.PlatyPS

Gijsreyn:

## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
N/A.

This pull request updates all Markdown files and external help file to use `Microsoft.PowerShell.PlatyPS`.


erwinvanhunen:
Hej Gijs! Thank you for this PR. It's HUGE! The next time can you please contact us before you do such a huge change (even if it's a minor change to each file, we still have to make sure everything works)? We're wondering what the reason is to change to the current prerelease version of Microsoft.PowerShell.PlatyPS? What are we gaining from it? The use case we have for PlatyPS is pretty limited and our documentation generation process currently works as intended. Performance is not really an issue either. I'd rather wait until the module has been released a full version instead of prerelease as things potentially can change in prereleases.

Gijsreyn:
Hey @erwinvanhunen. Thank you for your quick response. I understand. My apologies, I suppose in my excitement while creating an exercise for my upcoming book, I needed a live PS module with numerous Markdown files. After completing the exercise, I thought, why not commit all the files, even though I realized the implementation is quite limited. Next time, I won't hesitate to reach out to you (or other maintainers).

Perhaps I can still change your mind to take it into consideration in approving the pull request with some remarks (after it is tested ofcourse). If not, I can fully understand you abandon this PR. 

Here are my remarks:

- The new module catches out-of-date Markdown files that don't follow the expected elements (8 commands missed parameters)
- You mentioned it already, but indeed. Performance wise the `Microsoft.PowerShell.PlatyPS` is way faster
- So far the module produces a fraction more information than the old one, especially in the parameter section

The module is prerelease, but `PlatyPS` was never officially released to a stable v1 version and lacks the changes that are incorperated in `Microsoft.PowerShell.PlatyPS`. 

_A small side question. Do you still mind if I give it some love in my writing? :)_

erwinvanhunen:
Hej Gijs!

I checked the PR partially, but ran into issues with the new platyps. Some of the MD files in your PR generate errors. 

For instance Get-PnPContainerType.md: we noticed there is some mixup going on there between the ##parameters headers and the actual parameter declaration. 

Also we noticed in another cmdlet double ##EXAMPLES blocks (Get-PnPContainerTypeConfiguration.md). E.g. something is not 100% right in those files. So for now we're going to say no to this PR as it's simply too much work for us to work our way through all the MD files. 

However, we do plan to move to the new PlatyPS in the future, just not yet, as we also have to make sure that our site (pnp.github.io/powershell) will work as expected (we're using docfx there to generate the pages from the MD files, and if we have double example blocks we get some weird results too). Also we parse the frontmatters header in the MD files when we generate the docfx files which allows us to do some additional markup in the file (like stating that it's only available in the nightly build etc.). We have to make sure that this also still works. The new PlatyPS will help us there given that it's more object model based, so we don't have to have use our own parser there, but we're not there yet. 

Of course you can write about it in your upcoming book!

PR #4680, opened 2025-01-12
Refactor to handle AllowWebPropertyBagUpdateWhenDenyAddAndCustomizePagesIsEnabled 

reshmee011:
## Type ##
- [ ] Bug Fix

Refactor to check if AllowWebPropertyBagUpdateWhenDenyAddAndCustomizePagesIsEnabled is set before set the property bag without the enabling scripted at the site level.

Find more info on AllowWebPropertyBagUpdateWhenDenyAddAndCustomizePagesIsEnabled property from 
 https://learn.microsoft.com/en-us/answers/questions/1792615/setting-propertybag-property-without-allowing-cust


KoenZomers:
Thanks @reshmee011 !

PR #4705, opened 2025-01-26
Extended Restore-PnPRecycleBinItem functionality to restore selected set of items in bulk efficiently

namwar:
## Type ##
- [ ] Bug Fix
- [X ] New Feature
- [ ] Sample

Extended the functionality of **Restore-PnPRecycleBinItem** cmdlet by adding another parameter set. The new parameter set allows the user to provide a string array of Recycle Bin item IDs. It will enable users to deal with scenarios where many (but not all) items need to be restored. 
The caller can provide a list of IDs of items to bulk restore them. Bulk restore in this case will use REST API to restore items efficiently, in bulk. It also uses a fallback logic to deal with batch errors and automatically switches to individual restore, if needed.


src/Commands/Utilities/RecycleBinUtility.cs
```diff
@@ -99,8 +105,44 @@ internal static List<RecycleBinItemCollection> GetRecycleBinItemCollection(Clien
...
+            string idsString = string.Join("','", idsList); // Convert array to a comma-separated string  
+
+            try
+            {
+                string requestBody = $"{{'ids':['{idsString}']}}";
+                REST.RestHelper.Post(httpClient, apiCall,ctx, requestBody,"application/json", "application/json");
+            }
+            catch
+            {
+                //fall back logic
```
KoenZomers, line 127: Some verbose logging if this happens would be nice with the exception details so one can find out why it happens (i.e. too many items provided)

src/Commands/Utilities/RecycleBinUtility.cs
```diff
@@ -99,8 +105,44 @@ internal static List<RecycleBinItemCollection> GetRecycleBinItemCollection(Clien
...
+                    try
+                    {
+                        string requestBody = $"{{'ids':['{id}']}}";
+                        REST.RestHelper.Post(httpClient, apiCall, ctx, requestBody, "application/json", "application/json");
+
+                    }
+                    catch
+                    {
+                        //Digest errors because they can be due to to the following two reasons and we cannot do anything
+                        //1. Item with the same name already exists
```
KoenZomers, line 140: Also here it could be useful to add some verbose logging to give insights into the failures

KoenZomers:
Thanks @namwar for your efforts! Looks cool! Left two comments for you. Would appreciate it if you could have a look at those.

src/Commands/Utilities/RecycleBinUtility.cs
namwar: Thank you for your feedback. As suggested, I have added verbose logging in RecycleBinUtility.RestoreRecycleBinItemInBulk. Please have a look.

src/Commands/Utilities/RecycleBinUtility.cs
```diff
@@ -99,8 +105,61 @@ internal static List<RecycleBinItemCollection> GetRecycleBinItemCollection(Clien
...
+                            var odataError = e.Message;
+                            if (odataError != null)
+                            {
+                                if (odataError.Contains("Value does not fall within the expected range."))
+                                {
+                                    restoreRecycleBinItem.WriteVerbose($"Item - {id} already restored.");
+                                }
+                                else
+                                {
+                                    //Most common reason is that an item with the same name already exists. To restore the item, rename the existing item and try again
```
KoenZomers, line 154: Will the error that gets returned show the same info as you wrote in comments? That will be extremely helpful in identifying the cause of the issue. If the error itself already explains it, we're all good, if not, it could be worth addding this to the error message that gets shown.

namwar: Yes, the error message contains the exact reason of failure, as I mentioned in my comments. 
Following is an example error message from one of my test runs

**A file with this name \"Doc1.docx\" already exists in \"sites/DocTest/Shared Documents\". To restore the file, rename the existing file and try again.**

You can see that it provides the full information and resolution to resolve the problem.

Here is the full error message
{
  "odata.error": {
    "code": "-2147024816, Microsoft.SharePoint.SPException",
    "message": {
      "lang": "en-US",
      "value": "A file with this name \"Doc1.docx\" already exists in \"sites/DocTest/Shared Documents\". To restore the file, rename the existing file and try again."
    }
  }
}

namwar:
Please have a look at my comments on your feedback. I have clarified your concern. I think you can merge my code.

KoenZomers:
Looks good @namwar, thanks!

KoenZomers:
Still missing your change in the documentation by the way. Can you update Get-PnPRecycleBinItem to reflect this new option?

KoenZomers:
You also don't seem to have checked the option to allow us to update your PR branch so I can add the changelog entry to it. Can you enable that please?

namwar:
Thanks for your reply. I have checked and it is already checked for "Allow edits by maintainers". Am I missing something? Please guide me

image

namwar:
I will update the Get-PnPRecycleBinItem.md quickly

namwar:
I have updated the documentation/Restore-PnPRecycleBinItem.md. Please have a look.

namwar:
I had updated the documentation/Restore-PnPRecycleBinItem.md. Please have a look, when you can.

KoenZomers:
Sorry @namwar for taking so much time to get your PR merged. It's a rather complex one which stalls things from our side. If you're still interested in getting it merged, would you kindly have a look at the merge issues with the current dev branch and resolve these? Thanks!

KoenZomers:
Apologies @namwar this has taken forever. It's ready to be merged through https://github.com/pnp/powershell/pull/5199. Thanks for your contribution!

PR #4755, opened 2025-02-24
Updated `Reset-PnPDocumentID` to also support resetting the document id for a library

KoenZomers:
## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
N/A

`Reset-PnPDocumentID` now also supports resetting all document ids for all documents in a document library using the -Library parameter. This requires the -ContentType to be specified as well. It is not possible to run it on all files in a document library, regardless of their content type. This is a server side requirement. Notice that you need the content type applied to the document library itself, so i.e. not a 0x0101 but the specific content type of that on that library. See the sample in the documentation.


kachihro:
Thanks for this great new feature.

BUT - this doesn't work for DOCUMENT SETS - is that possible, somehow ??

PR #4811, opened 2025-03-25
Hotfix for Issue 4808

### LINKED ISSUE #4808: [BUG] Set-PnPListItem will get "Value cannot be null. (Parameter 'path')" error when you want to set taxonomy field with term id
### Reporting an Issue 
According to https://pnp.github.io/powershell/cmdlets/Set-PnPListItem.html, we can use term id

> Managed Metadata (single value with id of term): -Values @{"MetadataField" = "fe40a95b-2144-4fa2-b82a-0b3d0299d818"} with Id of term

to handle a taxonomy field (single one) update.
However I got **_"Value cannot be null. (Parameter 'path')"_** error when run `Set-PnPListItem`

### Expected behavior 
It should update the list item taxonomy field as expected

### Actual behavior 
It will throw **_"Value cannot be null. (Parameter 'path')"_** error 

Image

1.  Make sure you have a list which contains a taxonomy field (not allow multiple values)
2. You run the Set-PnPListItem command to update a item on this taxonomy field with a term id
like this
`
Set-PnPListItem -List "Demo List" -Identity $item -Values @{"Title" = "Test Title"; "taxonomyfield1"="47399476-176d-4cab-9596-4b74f7492740"}
`
Then you will get the error message.

(you can retrieve this by executing ```Get-Module -Name "PnP.PowerShell" -ListAvailable```)

I think I am using 2.12.4, actually I can reproduce the bug when I debug the dev branch source code

Image

- [x] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

wuxiaojun514:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #4808 

Currently `Set-PnPListItem` doesn't support term guid object for single taxonomy field, you have to convert it **ToString()** first, otherwise it cannot recognize it is guid and will try to run it as term path.

This is due to "`value as string`" cannot convert guid object to string, it will get null value. So I changed it to "`value?.ToString()`", the same one when handle multiple taxonomy field.


KoenZomers:
Thanks @wuxiaojun514 !

PR #4933, opened 2025-05-23
hotfix for #4932 to show better error message

### LINKED ISSUE #4932: [BUG]  It will throw "Get-PnPTerm: Value cannot be null. (Parameter 'clientObject')" error when try to get non-existent term with -Recursive mode
### Reporting an Issue 
I got **_Get-PnPTerm: Value cannot be null. (Parameter 'clientObject')_** error when I try to use `Get-PnPTerm` to get a wrong term (not existing one) when use Recursive switch. 

Like this command

`$fsTerm=Get-PnPTerm -Identity "wrong term name" -TermSet "My TermSet" -TermGroup "MyTermGroup"  -Recursive`

Image

If you don't use the **Recursive** switch, then command will tell you very clear that "The specified term doesn't exist"

Image

I think we should improve the error message

### Expected behavior 
Show "The specified term doesn't exist" error when used Recursive switch in `Get-PnPTerm` to get a wrong term

### Actual behavior 
It will show _**Get-PnPTerm: Value cannot be null. (Parameter 'clientObject')**_ error which is hard to understand the root cause

Run
`Get-PnPTerm -Identity "wrong term name" -TermSet "My TermSet" -TermGroup "MyTermGroup"  -Recursive`

3.1.0

- [x] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

wuxiaojun514:

## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #4932 

Right now if you want to use `Get-PnPTerm -Recursive`  command to get not-existing term, you will get the following error message, which is not clear enough.

> Get-PnPTerm: Value cannot be null. (Parameter 'clientObject')

The root casue is that if **term** is null, the  `clientContext.Load(term, expressions);` will throw this  **Value cannot be null. (Parameter 'clientObject')** error, so I added a check to better show the error message

I tested this hotfix with/without Recursive switch, it will always show "The specified term does not exist" error message


gautamdsheth:
Thanks @wuxiaojun514 , much appreciated the fix for this.

PR #5020, opened 2025-07-15
The AddDataRowsToSiteTemplate fields requested should also be validated if can be retrieved

PedroMordeP:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fix an issue, with a computed field type, "The property or field has not been initialized. It has not been requested or the request has not been executed. It may need to be explicitly requested" when executing [#L229](https://github.com/pnp/powershell/blob/3da15f2549a0a791043908f96c0887ebd8b47735/src/Commands/Provisioning/Site/AddDataRowsToSiteTemplate.cs#L229)

Improved AddDataRowsToSiteTemplate to validate also the "custom" Fields passed to the command.
Should also exclude the read-only and all unsupported Field Types.


gautamdsheth:
Thanks @PedroMordeP , much appreciated. Have merged the changes and they will be available tomorrow as well as the next major release !

Issue #5071, opened 2025-09-04
[FEATURE] Copy-PnPFileMetadata


alirobe:
I would like a cmdlet that synchronizes metadata between files and folders in SharePoint. Specifically, I want it to copy Created/Modified person and time. It can also optionally sync other metadata. 

This is to be run after a copy-pnpfile migration.

I have written a rudimentary version of this that suits my needs, so there will be an associated PR with this feature request.

gautamdsheth:
hey @alirobe , closing this as we merged your PR which is available in nightly builds and will be there in the major release !

PR #5072, opened 2025-09-04
Add Copy-PnPFileMetadata cmdlet and docs

alirobe:
## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
#5071 

 Copy-PnPFileMetadata cmdlet and docs

### DIFF
--- a/documentation/Copy-PnPFileMetadata.md
+++ b/documentation/Copy-PnPFileMetadata.md
@@ -1,0 +1,187 @@
+---
+Module Name: PnP.PowerShell
+schema: 2.0.0
+applicable: SharePoint Online
+online version: https://pnp.github.io/powershell/cmdlets/Copy-PnPFileMetadata.html
+external help file: PnP.PowerShell.dll-Help.xml
+title: Copy-PnPFileMetadata
+---
+  
+# Copy-PnPFileMetadata
+
+## SYNOPSIS
+
+Synchronizes metadata between files and folders in SharePoint
+
+## SYNTAX
+
+```powershell
+Copy-PnPFileMetadata [-SourceUrl] <String> [-TargetUrl] <String> [-Fields <String[]>] [-Recursive] [-Force] [-Connection <PnPConnection>] [-SourceConnection <PnPConnection>] [-TargetConnection <PnPConnection>]
+```
+
+## DESCRIPTION
+
+Synchronizes metadata (Created, Modified, Author, Editor) from source files and folders to their corresponding targets without copying the actual content. This cmdlet is particularly useful for restoring lost metadata after migration operations where system fields may have been reset.
+
+When updating items, the cmdlet uses `UpdateOverwriteVersion()` to allow setting system fields while avoiding new user-facing versions.
+
+For folder contents, files are processed one-by-one.
+
+## EXAMPLES
+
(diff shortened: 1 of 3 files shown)


alirobe:
Please ignore this for a few hours, I am working on some perf optimizations and simplifications. Will put another comment in when I am done.

alirobe:
OK, hopefully that's done. Ready for review :)

erwinvanhunen:
Very nice one! Thank you very much!

tborchardt:
Good morning team,

This is my first post in this GitHub repository, or github in general, so my apologies if i should post this comment as a new issue. I was just looking for confirmation if this cmdlet supports subsites because I'm having some issues getting it to copy metadata from one to a new site. Below is the code I'm running, with the domain name and client id redacted in the connection.

```
$src = Connect-PnPOnline -Url https://xyz.sharepoint.com/sites/storeengineering/retailfacilities -Interactive
$dst = Connect-PnPOnline -Url https://xyz.sharepoint.com/sites/retailfacilities -Interactive

Copy-PnPFileMetadata -SourceUrl "Shared Documents/Manuals" -TargetUrl "Shared Documents/Manuals" -SourceConnection $src -TargetConnection $dst -Verbose
```

When I attempt to run this, it comes back saying the source path '/sites/retailfacilities/shared documents/manuals' not found. 

I've tried changing the src connection to just https://xyz.sharepoint.com and then in the "Copy" command I used /sites/storeengineering/retailfacilities/Shared Documents/Manuals

Any guidance is greatly appreciated. If I can get this command to run, this will truly be a lifesaver for me as I work on promoting subsites to their own site.

alirobe:
Hmm.. Are you sure the path for the documents is actually 'Shared Documents' and not just 'Documents'?

Edit: on desktop now, and I see you're using a subsite. Didn't catch that at first, but for a subsite I think you should be connecting to the parent site and including the subsite in the path. Can you give that a try? @tborchardt 

I will review this later, confirm, and update docs if necessary.

tborchardt:
> Hmm.. Are you sure the path for the documents is actually 'Shared Documents' and not just 'Documents'?
> 
> Edit: on desktop now, and I see you're using a subsite. Didn't catch that at first, but for a subsite I think you should be connecting to the parent site and including the subsite in the path. Can you give that a try? @tborchardt
> 
> I will review this later, confirm, and update docs if necessary.

I was able to confirm that we need to connect to the parent site and then include the subsite in the path. 

It was able to successfully copy the default fields, except the user mapped ones. It says, "Applied non-user fields after user mapping failure." after each file it copied metadata for.

higgins0:
So happy you worked on this, being unable to achieve these fields update has been killing me in the last week using all sorts of context like PNP/CSOM. I know i can do it with REST but it's heavy (i'm a beginner). 
I'm also facing the  "Applied non-user fields after user mapping failure.". Does it have something to do with access to to user mapping information? Should we do something about the connection scope or credentials scope used?
[Update] eventually got the CSOM approach to work : was missing sufficient right on the sharepoint application associated to my user (my comment above led me to dig that way). Anyway, it would be so convenient if you can solve the is issue as that PNP cmdlet would be so much simpler to use...:)

PR #5096, opened 2025-09-26
Enhance Restore-PnPRecycleBinItem to Avoid List View Threshold Errors via Direct Item Access

irvcov:

## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

**Description**

This PR introduces a workaround to mitigate List View Threshold (LVT) errors encountered during recycle bin restore operations using Restore-PnPRecycleBinItem in SharePoint Online.
Previously, the cmdlet attempted to retrieve the recycle bin item via Identity.GetRecycleBinItem(Connection.PnPContext), which could trigger LVT violations when the recycle bin is large enough to exceed the list view limit.

image

This update adds logic to detect when Identity.Item is already populated and, in such cases, bypasses the call to GetRecycleBinItem. This allows the cmdlet to restore items directly from the provided object, avoiding enumeration of the entire recycle bin collection.

**Benefits**

Enables safe bulk restore operations using piped input:

**PowerShell> Get-PnPRecycleBinItem | Restore-PnPRecycleBinItemShow**

examples:

Get-PnPRecycleBinItem | Where-Object { $\_.Email -eq "johndoe@outlook.com" } | Restore-Pnprecyclebinitem -Force
Get-PnPRecycleBinItem | ? { $\_.Email -eq "johndoe@outlook.com" } | Restore-Pnprecyclebinitem -Force

- Avoids triggering LVT errors by skipping unnecessary collection access.
- Preserves existing behavior for single-item restores when Identity.Item is null.

**Limitations**

This is a client-side workaround; increasing or bypassing LVT constraints is not feasible as they are enforced server-side to protect backend performance.
The fix does not alter public API surface or expose internal identifiers.


gautamdsheth:
Thanks @irvcov , merged it. Much appreciated this fix ! It will be available in tomorrow's nightly builds as well as next major version when we release it.

irvcov:
@gautamdsheth Thanks for checking this! Just wanted to ask if these changes are already live, and if so, do you know from which version they're available?

PR #5134, opened 2025-10-25
Allow to call Register-PnPEntraIDApp with a certificate without private key

### LINKED ISSUE #5129: [FEATURE] Use Register-PnPEntraIdApp with CER file instead of PFX file
If I want to use an existing certificate when creating a new Entra App Registration, I need to pass in a path to the PFX file. But the PFX file should not actually be necessary. Entra ID expects a public key (a .CER file) when uploading a certificate file. The Private key is used in other places, in the case of PnP PowerShell it's used when signing in, for example when referencing the thumbprint or the path to the PFX itself.

The problem with passing in the PFX in the `Register-PnPEntraIdApp` command is that I don't want to. I may have installed the PFX in my windows credential store and thrown away the key (not being able to export it for security purposes, for example.) 

I'd like to be able to pass in the .CER file as a certificate (and not specify the password). PnP should just upload the file as a certificate and be done with it.

So now I'm forced to run

```ps
Register-PnPEntraIDApp -ApplicationName "some-name" -Tenant "contoso.onmicrosoft.com" -CertificatePath ".\some-cert.pfx" -CertificatePassword (ConvertTo-SecureString -String "somepassword" -AsPlainText)
```

but I essentially want to run:

```ps
Register-PnPEntraIDApp -ApplicationName "some-name" -Tenant "contoso.onmicrosoft.com" -CertificatePath ".\some-cert.cer"
```

jackpoz:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #5129

Allow to call Register-PnPEntraIDApp with a certificate without private key, so .cer instead of .pfx .
The private key is not needed when uploading a certificate to an app registration in Entra ID.

Here's the behavior with the PR:
image

The app registration works fine, see below a test case (notice how the pfx is used in this case with Connect-PnPOnline) :
image


NishkalankBezawada:
Hey thanks @jackpoz, Merged it. 

//Nish

PR #5147, opened 2025-11-01
update users in groups when parameter "-RemoveExisting" is used

svenboll:
## Type ##
- [X] Bug Fix
- [ ] New Feature
- [ ] Sample

I used this parameter in a project, and wondered why it didn't work.
I discoverd that the parameter "-RemoveExisting" i.e. in function "Add-Microsoft365GroupMember" (and others) was implemented, but completely out of function.

I want to give this PR into a little discussion, because i don't know if i really covered all scenarios in a good way.

Explanation:
While using Add-AzureADGroupe[Member|Owner] as i can see, the code split the input users into guid-users and none guid-users, and add them separately, so i am clearing the group before the both functions are called.
-> Clearing the Owners is not really possible, because there has to be one Owner left on the given group. So my PR could lead into having 2 owners on the group in the end, instead of the only one i want.

With M365Groups it was a bit easier, there i am using the "Update-[Members|Owners]" functions, which first adds not existing Users to the group and delete users, who should not be member of the group.

It is my first work on this project, and therefor i only want to make small changes and give you the possibility to dicuss. I don't want to really get deeper into huge changes, like combining addmembers and adddirectorymembers into one function (if it is even possible)

### Guidance ###
i was using version 3.1.0 of this module and you can reproduce it with any group


gautamdsheth:
Hey @svenboll , thank you and I welcome your contribution here 😊

Sorry, you had to run into this issue. 

I checked the history around this and looks like it never worked. And now, checking the docs, it might be a bit confusing as well. 

The best way forward, that I see, is that we should simply deprecate/obsolete the parameter in the cmdlets and also update the docs. 

We have Remove- Owner/Member etc cmdlets , so that covers the part of removing users from Groups. 

Could you please update the PR to make the parameters obsolete and update the docs about it ?

svenboll:
@gautamdsheth 
Thanks for answering to my first PR :-)

I will open a new PR when i go completely through all files. See you

PR #5213, opened 2026-01-19
Add -ClearSubScopes parameter to Set-PnPFolderPermission and Set-PnPListItemPermission. Closes #4951

### LINKED ISSUE #4951: [FEATURE] Set ClearSubscopes (BreakRoleInheritance ) for Permission Cmdlets (Set-PnPFolderPermissions and Set-PnPListItemPermissions)
I was surprised when I noticed that the the Cmdlets silently cleared all previously set permissions on all sub-folders.

I want to control whether subscopes should be cleared or not in the case permissions are not yet broken.

Check if HasUniqueRoleAssignments is false on $item, then use 
```
$item.BreakRoleInheritance($copyRoleAssignments, $clearSubscopes)
Invoke-PnPQuery
```
**Additional context**
C#
```
public virtual void BreakRoleInheritance(
    bool copyRoleAssignments,
    bool clearSubscopes
)
```

I checked commands that deal with permissions:

Set-PnPList
If BreakRoleInheritance used:
- CopyRoleAssignments: Default False
- ClearSubscopes: Default False

Set-PnPListPermission
This doesn't use BreakRoleInheritance on its own

Set-PnPFolderPermissions
BreakRoleInheritance is done when needed:
- ClearExisting: Default False, which translates to copyRoleAssignments: Default True
- ClearSubscopes: Always True / Not yet a parameter

Set-PnPListItemPermissions
BreakRoleInheritance is done when needed:
- ClearExisting: Default False, which translates to copyRoleAssignments: Default True
- ClearSubscopes: Always True / Not yet a parameter

**Implementation**

I can implement this myself,
but I am not sure if it is preferable to have in a non breaking way,
because in the current version clearSubscopes is always true.
Options:
Non breaking way

- Switch Parameter CopySubscopes or KeepSubscopes: Default False, which translates to clearSubscopes: Default True

The breaking way (because the hidden default is inverted)

- Switch Parameter ClearSubscopes: Default False

The other way

- Boolean Parameter ClearSubscopes: Default True

Any feedback is welcome! 😄

davidschenkUPG:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

## Related Issues? ##
Issue #4951

User can control whether subscopes should be cleared or not in the case permissions are not yet broken.
Preserves the original behavior of clearing subscopes, when the parameter is not specified.

Please let know, if you have improvements in of the wording in the Documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 - Added `Remove-PnPPowerAppPermission` cmdlet to remove Power Apps permissions. [#5168](https://github.com/pnp/powershell/pull/5168)
 - Added support for `-ClearGroupId` parameter in `Set-PnPSite` and `Set-PnPTenantSite` cmdlets. [#5192](https://github.com/pnp/powershell/pull/5192)
 - Added support for batch restoring items from the recycle bin using `Restore-PnPRecycleBinItem -ListIds` [#4705](https://github.com/pnp/powershell/pull/4705)
+- Added optional `-ClearSubScopes` flag to `Set-PnPFolderPermission` and `Set-PnPListItemPermission` cmdlets to clear unique permissions on child items/folders when breaking permission inheritance [#5213](https://github.com/pnp/powershell/pull/5213)
 
 ### Changed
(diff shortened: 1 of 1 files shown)


KoenZomers:
Note to self: add changelog entry

KoenZomers:
Thanks @davidschenkUPG for your contribution! Appreciate your efforts here. And welcome to the club!

Issue #3365, opened 2023-08-25
[BUG] Add-PnPListItem & Batching doesn't work with a French created site collection


hervedorierai3:
### Reporting an Issue or Missing Feature
I have to import data from CSV file to a list. I use Add-PnPListItem with -Batch option to improve the import time.

I developped my script in a English site collection : no problem at all.
I tested my script in a French site collection : the Add-PnPListItem fails with error "Object reference not set to an instance of an object."
If I remove the -Batch option, the items are created in the list (but it's very, very, very long).

I'm sure that it's not a permission issue because the script delete all item in the list (Remove-PnPListItem -Batch).
I'm connected with an Azure AD Application (Register-PnPAzureADApp -ApplicationName "MyBatchApp" -Tenant xxxx.onmicrosoft.com  -SharePointApplicationPermissions Sites.Manage.All -Password (ConvertTo-SecureString -String "xxxxxxx" -AsPlainText -Force) -DeviceLogin  )

### Expected behavior 
Would expect the item is successfully added to the list 😊

### Actual behavior 
Add-PnPListItem fails with error "Object reference not set to an instance of an object."

![image](https://github.com/pnp/powershell/assets/68116737/1c1ba761-ab1d-456b-a020-a2763cd95d87)

### Steps to reproduce behavior
- Create a "Communication" site collection with "French" as language.
- Create a simple list (only the Title column).
- Connect to the site collection with Certificate (Connect-PnPOnline -Url $siteurl -ClientId $application_clientid -Tenant $tenant_name -Thumbprint $certificate_thumbprint)
- Try to add an item to the list with Add-PnPListItem and -Batch option ==> Fails 😨
- Try to add an item to the list with Add-PnPListItem but without -Batch option ==> Success 😊

### What is the version of the Cmdlet module you are running?
2.2.55 nigthly

### Which operating system/environment are you running PnP PowerShell on?
- [ x] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

gautamdsheth:
@hervedorierai3 , please try to provide minimal steps to reproduce this issue.

Would suggest that you try it by adding a single list item via batch and also via CSOM and let us know.

hervedorierai3:
Added sample code.
Added mention that with interactive connection there is no problem.

hervedorierai3:
@gautamdsheth : 
Added sample code.
[...]

PR #5226, opened 2026-01-29
Include original exception information when throwing in `PnPConnectedCmdlet.ProcessRecord`

NoelTautges:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

When any cmdlet descending from `PnPConnectedCmdlet` encounters an exception while processing records (presumably, I'm not very familiar with all of this) that doesn't match its list of more specific exceptions, it throws a `PSInvalidOperationException`. [This behavior has previously been described as "error shallowing"](https://github.com/pnp/powershell/issues/3365#issuecomment-1709864014:~:text=Which%20correspond%20to%20this%20error%20shallowing%20code:) because it erases the more specific cause of the error, especially in errors with indistinctive messages like "Object reference not set to an instance of an object.".

This PR includes the original exception as an [inner exception](https://learn.microsoft.com/en-us/dotnet/api/system.management.automation.psinvalidoperationexception.-ctor?view=powershellsdk-7.4.0#system-management-automation-psinvalidoperationexception-ctor(system-string-system-exception)), which preserves that information when running `(Get-PnPException).Exception` and such.

If someone reviews this request- do you think it would be a good idea to add an inner exception element to `Get-PnPException`/`Get-PnPDiagnostics`? I'd be happy to make a PR for them too or add all that to this one 🙂

### DIFF
--- a/src/Commands/Base/PnPConnectedCmdlet.cs
+++ b/src/Commands/Base/PnPConnectedCmdlet.cs
@@ -101,7 +101,7 @@ namespace PnP.PowerShell.Commands.Base
                 // If the ErrorAction is not set to Stop, Ignore or SilentlyContinue throw an exception, otherwise just continue
                 if (!(new[] { "stop", "ignore", "silentlycontinue" }.Contains(ErrorActionSetting.ToLowerInvariant())))
                 {
-                    throw new PSInvalidOperationException(errorMessage);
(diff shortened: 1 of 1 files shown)


KoenZomers:
@NoelTautges Thanks for your contribution here! Appreciated. Please go ahead and file a PR for what you suggested.

PR #5230, opened 2026-02-03
New cmdlet Set-PnPFileVersion.Closes #5210

### LINKED ISSUE #5210: [FEATURE] Set-PnPListItemVersion
Hello,

I'm not sure if this is the right place.

I've enabled automatic version deletion at the tenant level, and after doing so, I have a client who regrets it because there are versions they don't want deleted.

Before the versions are deleted, SharePoint sets an expiration date, and I'd like to know if it's possible to remove that expiration date so that deletion is manual. I've been searching, but I can't find a cmdlet like **Set-PnPListItemVersion** to edit the file version properties, such as the expiration date.

Is there any way to do this? Do you know of a similar command? Could Get-PnPListItemVersion be implemented?

Thank you very much.

reshmee011:
## Type ##
- [ ] New Feature

## Related Issues? ##
Closes #5210.

New files
- documentation\Set-PnPFileVersion.md
- src\Commands\Lists\SetFileVersion.cs

Potentially needs to agree on name of cmdlet and location.


KoenZomers:
Please rename to `Set-PnPFileVersion` and add a changelog entry

PR #5254, opened 2026-02-26
Update code to use REST API instead of CSOM to solve the issue with root site. Closes #5249.

### LINKED ISSUE #5249: [BUG] Add-PnPNavigationNode behaviour issue on root site
Scripting the creation of hub navigation on the root site of SharePoint. 

`Add-PnPNavigationNode -Title "Test" -Url "https://mytenant.sharepoint.com/sites/targetsite" -Location TopNavigationBar`

This command on any other site under ...sharepoint.com/sites/<site> works fine

On the root SharePoint site it gives an error saying the target can't be found, it tells us it's the file '/sites/targetsite' that is missing. 

`Add-PnPNavigationNode: Cannot open "/sites/targetsite": no such file or folder.`

Adding the -external switch resolves this as it obviously stops it shortening the url to a relative one.

We have spent a week trying to work out what is going on here. The issue came up because the menu we were creating was a copy from a dev site where all the links were to "https://mytenant.sharepoint.com". These all created fine on the homesite but were being marked as ''internal' since we didn't use the switch. This meant that users were then unable to edit the target url in the GUI to another site or page. The GUI doesn't show the 'external' option and once a node is created without it there's no way to change it. The new GUI doesn't give any errors, it either undoes the change when you save or greys out the Save button and doesn't change it. Using the old GUI which you can still find on the Site Settings page shows the same error, Cannot open "/sites/targetsite": no such file or folder.

What is the issue? It may be a SharePoint architecture issue rather than PnP but some thoughts

- Can -external be on by default. The property appears to be deprecated by MS, in the GUI its not an option so having it set automatically will remove this issue?
- Why can't the target be found, the relative url looks right  '/sites/targetsite' ?
- If I'm giving the full url and not /sites/targetsite should it not be able to work out what to do slightly better?

3.1.0

reshmee011:
Update code to use REST API instead of CSOM to solve the issue with root site. Closes #5249.

## Type ##
- [ ] Bug Fix

## Related Issues? ##
Fixes #5249.

### DIFF
--- a/src/Commands/Navigation/AddNavigationNode.cs
+++ b/src/Commands/Navigation/AddNavigationNode.cs
@@ -43,11 +43,6 @@ namespace PnP.PowerShell.Commands.Branding
         [Parameter(ParameterSetName = ParameterSet_Default)]
         [Parameter(ParameterSetName = ParameterSet_PreviousNode)]
         [Parameter(Mandatory = false)]
-        public SwitchParameter External;
-
-        [Parameter(ParameterSetName = ParameterSet_Default)]
-        [Parameter(ParameterSetName = ParameterSet_PreviousNode)]
-        [Parameter(Mandatory = false)]
         public List<Guid> AudienceIds;
 
         [Parameter(ParameterSetName = ParameterSet_Default)]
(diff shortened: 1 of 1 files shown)


gautamdsheth:
Hey @reshmee011 , not sure if we can merge it in this release. It is a breaking change as far as I can see. We need to return a CSOM object here. Maybe we can do it in next major release like v4 ?

reshmee011:
The cmdlet behaves the same as it did when we combined CSOM and REST, just with one fewer API call now. The only change is that the REST endpoint no longer exposes the external flag, but that doesn’t affect functionality I can still add external links like http://www.google.com without it and we could mark `external` parameter as deprecated.

gautamdsheth:
It is a breaking change because the output type has changed here. 

1) Currently implementation outputs `Microsoft.SharePoint.Client.NavigationNode` whereas this PR outputs `Model.SharePoint.NavigationNode`. This change should only be done a major release. 

2) The IsExternal parameter should be deprecated if we dont use it. 

3) We also need to check if this also works on old classic SP sites, not just modern sites.

reshmee011:
@gautamdsheth : Good points ! Agree let's postpone it for next release.

PR #5358, opened 2026-06-22
Add support for applying built-in Microsoft site designs (store 1)

fabianhutzli:
## Type ##
- [ ] Bug Fix
- [x] New Feature
- [ ] Sample

## Summary

Adds support for Microsoft's built-in SharePoint site designs (the "store 1" designs) to two existing cmdlets. Previously, `Get-PnPSiteDesign` and `Invoke-PnPSiteDesign` only worked with tenant-registered custom designs via CSOM. Built-in Microsoft site templates such as **Event**, **Department** and **Human Resources** live in a separate store and can only be reached via the `SiteScriptUtility` REST API with `"store": 1` — making them inaccessible from PnP PowerShell without falling back to raw `Invoke-PnPSPRestMethod` calls.

## Modified cmdlets

**`Get-PnPSiteDesign`**
- New `-BuiltIn` switch (new parameter set). When specified, calls `SiteScriptUtility.GetSiteDesigns` with `store: 1` and returns all built-in Microsoft site designs with their `Id`, `Title` and resolved `Template` enum value.
- Existing default behaviour (listing and looking up tenant-custom designs) is unchanged.

**`Invoke-PnPSiteDesign`**
- New `-Template <BuiltInSiteTemplates>` parameter (new parameter set). Resolves the GUID from the existing `BuiltInSiteTemplates` enum mapping and calls `SiteScriptUtility.ApplySiteDesign` with `store: 1`.
- Existing `-Identity` parameter set (CSOM path for custom designs) is unchanged.

## Implementation notes

- The `BuiltInSiteTemplates` enum and its GUID mappings in `BuiltInSiteTemplateSettings` already existed in the codebase (used by `Get/Set-PnPBuiltInSiteTemplateSettings`). The new code reuses these mappings directly rather than introducing any new GUIDs.
- Two utility methods were added to `SiteTemplates.cs`: `GetBuiltInSiteDesigns` and `ApplyBuiltInSiteDesign`. Both POST to the `SiteScriptUtility` REST endpoint with `"store": 1`.
- A `BuiltInSiteDesign` model class was added to deserialise the REST response from `GetSiteDesigns` (store 1), which returns `Id` and `Title` rather than the full CSOM `TenantSiteDesign` object.
- `Get-PnPSiteDesign` was missing a `DefaultParameterSetName`, which would have caused an ambiguous parameter set error when called without arguments after the second parameter set was added. Fixed as part of this PR.
- Documentation included
- Developed with AI assistance
- Tested


gautamdsheth:
Thanks @fabianhutzli, merged it , much appreciated !!!

PR #5373, opened 2026-06-29
Add the ability to create an explicit crawled property at the tenant level

pschneid:
## Type ##
- [ ] Bug Fix
- [X] New Feature
- [ ] Sample

## Related Issues? ##
N/A

This utilizes the ability to import a search configuration to allow users to create explicit crawled properties at the tenant level.  I've limited this to the tenant level because if you create one at the site level you will always have to maintain the mappings separately from the tenant-level.

The preferred usage is to use the -PropSet flag and choose the appropriate choice.  It will warn you if you try to create an ows_q_XYZ property using the taxonomy property set, for example.  People with need can use the -PropSetGuid option to create rare or non-standard property sets (not recommended).

There are times where SharePoint admins may, for example, need to make a crawled property explicit but can't see it on the schema admin page because, for example, they don't have access to any content that has an implicit version of the property.  This allows them to work around that issue.


pschneid:
- Switched to the existing Resources alias for the confirmation prompt.
- Cleaned up the output object initializer and made CategoryName/MapToContents explicit.
- Added an assertion for the reflected test helper lookup.
- Reordered the parameter docs alphabetically.
- Added the standard Related Links section.

pschneid:
Addressed latest feedback.

wobba:
@erwinvanhunen this one looks good to me.

KoenZomers:
@pschneid @wobba Any particular reason for not having the cmdlet inherit from PnPAdminCmdlet but instead building in manual validation based on -admin in the URL?

pschneid:
> @pschneid @wobba Any particular reason for not having the cmdlet inherit from PnPAdminCmdlet but instead building in manual validation based on -admin in the URL?

Because I'm not an expert in PnP and am trying to help add something that will make people's lives easier?

wobba:
> @pschneid @wobba Any particular reason for not having the cmdlet inherit from PnPAdminCmdlet but instead building in manual validation based on -admin in the URL?

@pschneid please make the suggested change. Inheriting from admin makes the admin connection better/easier.

pschneid:
Switched to using admin context.

KoenZomers:
Thanks @pschneid . Will test it out here as well.

KoenZomers:
Thanks @pschneid for your valuable contribution! Keep up the good work!

Issue #5103, opened 2025-09-30
[FEATURE] Support for Microsoft 365 Multi-Geo cmdlets


Borgquite:
**Is your feature request related to a problem? Please describe.**
At present it is not possible to use various Multi-Geo cmdlets inside PnP-PowerShell. For example, there is no equivalent for:
Get-SPOGeoMoveCrossCompatibilityStatus
Get-SPOUserAndContentMoveState
Start-SPOUserAndContentMove
Get-SPOSiteContentMoveState
Start-SPOSiteContentMove
Get-SPOUnifiedGroup
Set-SPOUnifiedGroup
Get-SPOUnifiedGroupMoveState
Start-SPOUnifiedGroupMove
Set-SPOMultiGeoExperience

These are all available in the Microsoft SPO Management Shell, but due to the limitations of this module, there appears to currently be no way to perform these MultiGeo functions e.g. via an app registration, or using a managed identity.

**Describe the solution you'd like**
PnP PowerShell cmdlets to be implemented to migrate OneDrive, SharePoint sites and Microsoft 365 Groups using MultiGeo functionality

**Describe alternatives you've considered**
It is possible to manage these via the Microsoft SPO Management Shell but this does not support app registrations or managed identity, and so unattended scripts must use basic authentication(!)

**Additional context**
See discussion here: https://github.com/pnp/powershell/discussions/3536

gibi916:
+1 Also need it. Thanks !

KoenZomers:
We can look into adding these. Challenge we have as a team is that we do not have access to a multi geo tenant to test these against. If we add them, can you test them for us? 

Alternatively, the spo management module should now support unattended authentication as well. This has recently been added. Not sure about managed identity support. Worth looking into as well.

Borgquite:
@KoenZomers Yes, I can do a bit of light testing against the nightly build if that would help!

KoenZomers:
It would be a heck of a job to create all these manually. I gave it a shot using AI based on the SPO Management Module in this PR: https://github.com/pnp/powershell/pull/5215

@Borgquite, are you able to build a version out of this PR and test them against your multi-geo tenant? Fingers crossed that AI did a good job here.

KoenZomers:
Looks like it generated stubs for at least, potentially all, cmdlets :( So it will not return any actual data likely. Give them a try anyway and let me know which of these is the most important cmdlet for you. I can check if I can easily fix it, but afraid it will be a very (too) time consuming task.

[...]

PR #5401, opened 2026-07-18
Fix multi-geo SPO compatibility issues

gautamdsheth:

## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Addresses feedback in #5103.

Fixes multi-geo compatibility issues reported against the SPO Management Shell behavior:

- Reads `Get-PnPGeoAdministrator` results as an OData feed so `value`/array responses deserialize correctly.
- Aligns `GeoAdministrator` output property order with the SPO Management Shell type.
- Removes the extra generic `ShouldProcess` prompt from `Set-PnPMultiGeoExperience`, leaving the single SPO-style irreversible-upgrade confirmation.
- Aligns the unsupported-version error argument for `Set-PnPMultiGeoCompanyAllowedDataLocation` with the rest of the MultiGeo client.
- Updates the relevant cmdlet documentation and changelog entry.


gautamdsheth:
Addressed the Set-PnPMultiGeoExperience review feedback in dd5be705e. Normal execution still matches SPO Management Shell by using only the single irreversible-upgrade ShouldContinue prompt, while -WhatIf now returns before the support check/prompt/upgrade by calling ShouldProcess only for the WhatIf path.

PR #5412, opened 2026-07-24
Fix Set-PnPDefaultColumnValues corrupting library defaults on invalid taxonomy values

### LINKED ISSUE #5409: Set-PnPDefaultColumnValues Index was outside the bounds of the array
Error when I attempt to set the default column for text field or managed metadata I get an error:

### Expected behavior 
I expect the default column value for the field to be set. Just as in the case of setting and inspecting via the GUI
Setting a managed metadata column default in any other way other than term id break the document library defaulting mechanism for the column in question.

### Actual behavior 
Please describe what you see instead. Please provide samples of output or screenshots.

Set-pnpDefaultColumnValues -List "Documents" -Field "mmdCol" -Value "term1"
Set-PnPDefaultColumnValues: Index was outside the bounds of the array.

-Create site
-add (in my case via contenthub) content type with managed metadata column (not required, single value, no translations/properties or anything)

-enable content types on doc library

---------
All through the sequence of testing and reproducing I am testing if defaults are working in general via a command such as:
Set-pnpDefaultColumnValues -List "Docs" -Field "City" -Value "Stoke" is working still...
----------

-When I attempt to set the mmd column:
Set-pnpDefaultColumnValues -List "Docs" -Field "Programme" -Value "myCompany|Programme|WrongValue"
or 
Set-pnpDefaultColumnValues -List "Docs" -Field "Programme" -Value "myCompany|Programme|CorrectValue"
or 
Set-pnpDefaultColumnValues -List "Docs" -Field "Programme" -Value "CorrectValue"

The command takes it - it appears not to check at some level.

Any of these means of specifying the term causes some configuration error in the document library and renders even the GUI incapable of correcting the default value. Gui shows an error as well. when attempting to configure that a general site error occurs.
even removing the column and content type, rolling back to original settings yields, even for my test 'city' default setting of 'Set-PnPDefaultColumnValues: Index was outside the bounds of the array.'

The only way to fix the library I found was to delete it.

What works?

Setting via the termId in the 

$myTermId = (Get-PnPTerm -TermSet "My Term Set" -TermGroup "My Term Group" -Identity "TERM1").id
Set-PnPDefaultColumnValues -List "Documents" -Field "mmdColumn" -Value $myTermId

3.3.0

- [X] Windows
- [ ] Linux
- [ ] MacOS
- [ ] Azure Cloud Shell
- [ ] Azure Functions
- [ ] Other : please specify

svermaak:
## Type of change

- [x] Bug fix

## Related issue

Fixes #5409

## What is in this Pull Request

When `Set-PnPDefaultColumnValues` is used on a managed metadata field with a value that cannot be resolved to a term (a bare label like `ReproTerm`, or a wrong path), the unresolvable value is silently dropped by PnP Framework's `GetDefaultColumnValueFromField` and a malformed entry is written to the library's `Forms/client_LocationBasedDefaults.html`. From that moment on, every default column value operation on that library fails with "Index was outside the bounds of the array" - for any field, including plain text fields - because the framework's parser indexes into the malformed entry without bounds checks. The library stays broken until the defaults file is deleted.

This PR validates taxonomy values up front in the cmdlet: each value must resolve to a term, either by term id or by full `TermGroup|TermSet|Term` path. If a value does not resolve, a clear error is thrown and nothing is written:

> Value 'ReproTerm' could not be resolved to a term. Provide a term id or the full path to the term in the format 'TermGroup|TermSet|Term'. The default column values on the list have not been changed.

Valid term ids and paths behave exactly as before.

## Reproduction and verification

Reproduced and verified against a SharePoint Online test site with a document library containing a text column and a taxonomy column.

### Before (v3.2.0): invalid value accepted silently, library defaults permanently broken

Before the fix

### After (this fix): clear error, nothing written, library stays healthy, valid values still work

After the fix

## Workaround for already broken libraries

Deleting `<library>/Forms/client_LocationBasedDefaults.html` (e.g. via `Remove-PnPFile`) restores default column value functionality without recreating the library.

Note: the unguarded parsing of existing entries lives in PnP Framework's `SetDefaultColumnValues`/`GetDefaultColumnValues` extensions; hardening that parser so a corrupted file recovers gracefully would be a follow-up in the pnpframework repo.


gautamdsheth:
Thanks @svermaak , merged this !
Much appreciated

PR #5468, opened 2026-09-10
Fix #5467: related to default column value / event receiver issue

### LINKED ISSUE #5467: [BUG] Set-PnPDefaultColumnValues: Error applying default column values - writes unresolved WssId -1 for taxonomy fields on newly-created site collections (regression — worked last week)
### Environment
PnP.PowerShell: 3.4.1 (reinstalled/updated during investigation, issue persists on both old and new versions)
Windows 11
Reproduced on two separate tenants (one DEV and one PROD in French zone) as a SharePoint admin + site collection admin on target sites.

### Summary
Set-PnPDefaultColumnValues against a managed-metadata (TaxonomyFieldType) column throws a generic "Error applying default column values" (no inner exception, no stack trace) on any newly-created site collection. The command still partially writes client_LocationBasedDefaults.html, but with WssId = -1 instead of a resolved value, e.g.:
`-1;#CONTOSO - S0072|f5722078-1dc0-4adb-8001-f781bb7a1fec`
A -1 WssId does not get applied to new documents on upload — Library Settings shows the default as "set," but dropped files don't receive it. Manually clearing and re-setting the same default via the browser UI resolves it to the correct WssId (e.g. 4;#...) and it then works normally.
I create and provision site nearly every week, I was working last week.

###  Key evidence this is a regression on new site collections specifically, probably on MS side, not a script/template bug

The exact same provisioning script and PnP template that fail today provisioned a working site last week.
Recreating that same, previously-working site from scratch reproduces the failure on two independent tenants.
An older, long-lived site collection with an identical -1;#Label|GUID string in its client_LocationBasedDefaults.html does apply defaults correctly on upload -> implying resolution used to succeed at write time (or via some now-changed mechanism) and no longer does on freshly created sites.
The term is confirmed present with a valid WssId in the site's TaxonomyHiddenList before the call — so it isn't a case of the term never having been used before; the cmdlet still writes -1 regardless.

###  Repro steps
Create a new SharePoint Online site collection.
Provision a document library with a TaxonomyFieldType site column bound to an existing term set.
Ensure the term has already been assigned to at least one item on the site (rules out "unused term" explanations).
Run: 
`Set-PnPDefaultColumnValues -List '<library>' -Field '<taxonomy field internal name>' -Value '<TermGroup>|<TermSet>|<TermLabel>'`
Observe the generic error, and inspect <library root>/Forms/client_LocationBasedDefaults.html : the WssId is -1 and default value is not applied when adding a new file (drag and drop).
Expected: default value is not applied when adding a new file (drag and drop) and/or cmdlet resolves and writes the term's actual WssId, or throws an error with the real underlying exception surfaced.
Actual: generic error with no detail, no default value applied and a non-functional -1 written to the defaults file.
Note : I also tried on brand new doclib with simple name "TEST", also tried passing directly the term GUID, same behavior.

gautamdsheth:
## Type ##
- [x] Bug Fix
- [ ] New Feature
- [ ] Sample

## Related Issues? ##
Fixes #5467

Log warning for event receiver issue when setting default column value

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Added
 
 ### Changed
-- `Set-PnPDefaultColumnValues` no longer fails with a generic 'Error applying default column values' when SharePoint Online rejects registering the LocationBasedMetadataDefaultsReceiver event receiver on the library, which it does since September 2026. The values are now stored and a warning explains that SharePoint will not apply them to newly added items until the receiver can be registered. Requires the accompanying PnP Framework change. [#5467](https://github.com/pnp/powershell/issues/5467)
 - Telemetry in PnP PowerShell has been removed due to the costs of collecting the data didn't outweigh the benefits to the PnP PowerShell team to have insights into its usage. The involved cmdlets `Get-PnPPowerShellTelemetryEnabled`, `Enable-PnPPowerShellTelemetry` and `Disable-PnPPowerShellTelemetry` have been marked as deprecated and no longer function, but will stay in v3 for backwards compatibility with existing scripts. These cmdlets will be removed in the next v4 release. All versions of PnP PowerShell will no longer be able to submit telemetry. You might see background requests for this failing. This will not interfear with the normal execution of your PowerShell script. [#5460](https://github.com/pnp/powershell/pull/5460)
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


mdissel:
For now we found an alternative: 
Post the registration of the event receiver to the SharePoint REST API via 
`$"{webUrl.TrimEnd('/')}/_api/web/lists(guid'{listId}')/EventReceivers"`

mdissel:
In my opinion this is not a fix, it's a bypass.

If we don't want to use the rest api as an alternative (no idea how long that keeps working) my suggestion would to let it fail by default and implement another book parameter skipAddingEventReciever

